Issue #195, opened 2018-04-23
Blokada is using a lot of background data even when background data is disabled for the app


bjwessels:
### Steps to reproduce
1. I just have blokada running
2. Background data for blokada has been turned off on android
3.

### Expected behaviour

No Background data should be used - I checked and every day I see
This started to happen on the 21st April. It uses between 2-6 Mb of background data
I think this has only started when I upgraded recently  

### Actual behaviour
No background data should be used 

### Configuration
1. Blokada version 3.3

Keep alive is on
start on boot is on
online only is on

2. Device model
Moto E4

3. Android version / ROM
Android 7.1.1

### Questions
1. How long have you been using Blokada? 
Several months

2. How long time ago has the problem started to happen? 
The 21st April

3.  How often does this problem happen to you?
Every day there is background usage 2 to 6 Mb a day
![screenshot_20180423-124736](https://user-images.githubusercontent.com/3920936/39141109-b8a0af36-46f4-11e8-8f68-d2ba24dcbe00.png)

kar:
Can anyone link to an authoritative definition of "background data" in Android? I can't guess what it's counting, but basically data usage in Blokada is currently:

- Fetching filter configuration / defaults from blokada.org (tiny amount of text, roughly rechecked once a day)
- Fetching actual filters (much more text, depending on your filters choice), fetched roughly once a day  - this changed recently. used to be once a week, but people were unhappy they don't get newest filters. Also your actions (changing filter config) will cause refetch so it may add up - but this probably won't count as background data
- ACRA crash reports - sent only if you explicitly accept it (a dialog shows up after crash)

bjwessels:
Just to clarify - I checked the usage history and it's once a day and about 3 Megs. Today (23/4/2018) there were 2 times each of 3 Meg. Going back further I see it was using a tiny amount < 0.1 Meg if at all.

I have a limited data plan

kar:
which lists do you use?

kar:
Thanks. Also may be related, there's a feature request to provide control over how often lists are refreshed, this could help your case.

bjwessels:
Ok so using the default list AdZhost only and keep alive = on , online only = on
All times EST 
27/04/18 = 8am - 1.75b Mb 11am - 1.78Mb  3pm - 1.78,  total = 5.32 Mb of background data
26/04/18 = 8am - 1.78 Mb
25/04/18 = 8am - 1.78 Mb

[...]

PR #214, opened 2018-05-19
Set cache update timeout to one day

aerusso:
The existing definition of cacheTTLMillis was specified in centiseconds, rather than milliseconds, leading to a cache file update 10 times per day. This updates that to be once a day, which should be sufficient for most people, at least until a tunable parameter can be specified.

This 10x daily refreshing was responsible for some ridiculous data usage. Although not directly the issue for, e.g., #195, this certainly exacerbates those situations.


kar:
Nice catch!

Issue #94, opened 2018-02-01
Ignores IPv6 DNS results


gloomytrousers:
I'm on an IPv6-capable (dual stack) network. With Blokada running, it seems no IPv6 (AAAA) records are returned in DNS results, only IPv4 (A) records. This makes IPv6-only sites inaccessible, and forces access to dual-stack sites via IPv4 only.

This can be tested by visiting an IPv6 test site with and without Blokada running.

Mikaela:
I am behing carrier grade NAT (and most of mobile IPv6 users are likely in similar situation with DualStack Lite) also at home and lately it has seemed like either my router or the CGN is overloaded and my IPv4 connections are very slow or get cut or otherwise have issues while IPv6 works just fine.

Thus I think this issue is blocking me from using Blokada.

cbruegg:
I've (hopefully) resolved this issue in PR #295 :smiley:

cbruegg:
The PR for this is now merged, so this issue should be fixed in the next release.

colbycdev:
Issue unresolved in v3.7, which in the changelog says adds ipv6 support. I ping my (ipv6 ready) server IP, and it informs me 'no route to host'. Turn blokada off, and works as expected. The site ipv6-test.com also fails to see my ipv6 address I know I have. It says I'm not ipv6 capable (tried the site in both chrome and Firefox).
I'm running Android 8.1, on an Alcatel Tetra (soon to be Moto g6 later this week. Yay!)

cbruegg:
My patch was reverted unfortunately, but I don't know why. If somebody
tells me the reason, I would love to look into it.

colbycdev <notifications@github.com> schrieb am Mo., 4. März 2019, 23:03:

InsaneSplash:
I hope this is resolved soon.

rpeter85:
It was reverted because this part of the code had a countereffect on the DNS fallback option:
```
if (dnsServers.none { it.getAddress() is Inet6Address }) {
 // If we add no IPv6 route or DNS server, all IPv6 access is otherwise blocked by default
 builder.allowFamily(OsConstants.AF_INET6)
}
```
Unfortunately it blocked the connection on devices where we faced #284

cbruegg:
@rpeter85 Thanks for the information! Could you please elaborate how this interferes with the DNS fallback? All it should do is allow IPv6 traffic _in addition_ to IPv4 traffic.

colbycdev:
[...]

PR #295, opened 2018-11-23
Enable basic IPv6 support - DNS results now include AAAA records

cbruegg:
This should resolve issue #94. The issue was that when the `VpnConfigure.Builder` is not given any IPv6 address as a `addRoute` or `addDnsServer` call, it blocks all IPv6 communication by default. This should not break the blocking functionality since dnsjava supports AAAA queries as well and the `question.name` (see `Proxy.kt:55`) remains included in any case, which allows Blokada to block any requests to disallowed hostnames.

### DIFF
--- a/app/src/tunnel/kotlin/tunnel/VpnConfigurator.kt
+++ b/app/src/tunnel/kotlin/tunnel/VpnConfigurator.kt
@@ -61,7 +61,7 @@ internal class VpnConfigurator(
             }
         }
         if (dnsServers.none { it is Inet6Address }) {
-            // If we add no IPv6 route or DNS server, all IPv6 access is blocked by default
+            // If we add no IPv6 route or DNS server, all IPv6 access is otherwise blocked by default
             builder.allowFamily(OsConstants.AF_INET6)
         }
 
--- a/app/src/tunnel/kotlin/tunnel/VpnConfigurator.kt
+++ b/app/src/tunnel/kotlin/tunnel/VpnConfigurator.kt
@@ -1,6 +1,7 @@
 package tunnel
 
 import android.net.VpnService
+import android.system.OsConstants
 import core.Kontext
 import core.Result
 import java.net.Inet4Address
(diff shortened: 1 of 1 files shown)


kar:
@cbruegg Good idea, thanks. Just to be sure, have you tried to run a build with this change on your device yet?

cbruegg:
Yes, I've been running a build with this change for a few weeks now. Works well.

colbycdev:
Out of curiosity, how'd you get the build? What sort of shell command is needed on a Ubuntu machine?
________________________________

cbruegg:
@colbycdev I ran it straight from Android Studio. I suppose you can achieve the same result by running Make using the [Makefile here](https://github.com/blokadaorg/blokada/blob/master/Makefile).

colbycdev:
@cbruegg I can't build. Some software property concerning a key is causing gradle sync to fail in Android Studio on Ubuntu 18.10

_Sent from my TETRA using [FastHub-Libre](https://github.com/thermatk/FastHub-Libre/)_

cbruegg:
@colbycdev Ah yes. You'll need to build using `Build/Generate Signed Bundle or APK`. Select _APK_, then create a signing key. Select build variant `adblockerFdroidRelease` (or at least that's the one I chose) and sign with V1 and V2. After clicking on _Finish_, the build should run and after a while a message should appear in the bottom right corner that it's complete.

kar:
@colbycdev just choose adblockerHomeDebug flavor.

colbycdev:
First, I'm never given that choice @kar 
Second, here's a screenshot of it
https://drive.google.com/file/d/144roMzrTD8wmOxSTyV4xv0xo-ql7o83n/view?usp=drivesdk

colbycdev:
This claims to fix IPv6, yet in the version 3.7 which claims to have IPv6 support, IPv6 doesn't work? At least remove the entry from the changelog

PR #305, opened 2018-12-15
Rename About us.md to about_us.md

pan93412:
Use space in filename isn't a good idea ;)


animeking9901:
Traditionally terminals don't like spaces. Good spot

PR #374, opened 2019-04-23
Update App.kt

Thomas499:
Most users will not want a new notification each time something is blocked. This will annoy the user. By default have notifications off. Permanent notifications are ok because they don't give user impression they have a new message each time something is blocked. But new notifications are not wanted.

### DIFF
--- a/app/src/legacy/kotlin/core/App.kt
+++ b/app/src/legacy/kotlin/core/App.kt
@@ -34,7 +34,7 @@ class AUiState(
     )
 
     override val notifications = newPersistedProperty(kctx, APrefsPersistence(ctx, "notifications"),
-            { true }
+            { false } // By default, have notifications off. 
     )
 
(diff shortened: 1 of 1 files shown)


kar:
I'm not sure about that. The reason is that new users will see "something is working", otherwise they may wonder if anything is happening at all. What others think?

ps100000:
I don't think it's a problem to have notifications enabled by default. I mean, it's as easy as opening the app and clicking a button if you're annoyed that Blokada sends you notifications. I don't know of any other app that has this setting so easy to find. I think it's important to give a new user a real time indicator that it's working.  If you want to improve this, maybe add a simple dialog that opens shortly after the "hello message", sends a "sample notification" and asks the user if he likes it or not?

kar:
We discussed it on Telegram, the general consensus seems to be to turn them off by default. Later on we'll have an onboarding wizard so then we can let the user choose.

PR #414, opened 2019-06-27
Fix f-droid link

ribtoks:


PrintableCharacter:
Could you specify a little more?
Do you want to tell that there's only v3 available on Fdroid? The v4 update will probably be added to Fdroid next weeks to make sure that Fdroid users get a version which is as stable as possible.

ribtoks:
@PrintableCharacter Did you take a look into a diff?

PR #428, opened 2019-07-09
Added custom names for custom DNS server

ps100000:
Added another step for creating a custom DNS where the user can set a name for the new entry.


ps100000:
The commits of this PR are already merged so it's not needed anymore.

PR #436, opened 2019-07-17
Added button for clearing the live history/host log

### LINKED ISSUE #413: I want to clear my Host log.
I'd like to be able to clear my Host Log from time to time, for example to show the App to my friends safely.

ps100000:
Fixes #413

### DIFF
--- a/app/src/legacy/res/values/strings_bits.xml
+++ b/app/src/legacy/res/values/strings_bits.xml
@@ -46,6 +46,7 @@
     <string name="menu_ads">Ads</string>
 
     <string name="menu_ads_live_label">Live history of allowed and blocked requests</string>
+    <string name="menu_ads_clear_log">Clear live history</string>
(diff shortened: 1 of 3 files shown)


app/src/legacy/res/values/strings_bits.xml
```diff
@@ -46,6 +46,7 @@
     <string name="menu_ads">Ads</string>
 
     <string name="menu_ads_live_label">Live history of allowed and blocked requests</string>
+    <string name="menu_ads_clear_log">Clear live history</string>
```
kar: Change "live history" to "host log" as we use it elsewhere

kar:
Actually there's a bug.

1. clear log
2. close menu
3. do sth to generate new activity in log
4. go back to log
5. search and clear button not visible
6. navigate back and enter the log again
7. search and clear button visible

Probably something about not preserving those first two items as special case.

PR #479, opened 2019-08-14
Fix typo

dandv:


kar:
Things been moved around since then, but I applied it manually. thank you!

PR #481, opened 2019-08-15
Changed widget to show numbers between 100k and 1m correctly

ps100000:


speedlever:
Is there room in the field to show the actual number of blocks instead of a decimal rounded number? Right now my widget just shows 0.2m and I assume it will show that number for actual numbers between 150,000 and 249,999. My actual number is presently 187,xxx.

Issue #198, opened 2018-04-26
Support DNS-over-TLS / DNS-over-HTTPS


jeroenev:
With more dns servers like Cloudflares supporting encrypted DNS, would be nice if that could be supported through blokada? Not sure how doable it is/how much blokada relies on android's internal logic for DNS requests, but it would be nice if it would work, not everyone has a rooted device with support for DNScrypt

StarTrekRocks73:
Hi. I completely agree with you. DNS requests are normally not encrypted meaning that they can be tracked or redirected to a dangerous site. Encrypting the traffic through HTTPS/DNSCrypt protocol would be a major improvement on both a privacy and security.

late-latte:
Seconded! I live in a country where the government is poisoning our DNS. I'd really appreciate it if Blokada somehow can listen to non-53 port. It'll help unrooted devices poised to not receive Android P update a lot. I once asked DNS66 dev to do it and the dev seemed to think it's achievable, although the project now seems to hit a wall.

kirbyfan64:
FWIW Android P is coming with DNS over HTTPS built-in, but I'm not sure if Blokada messes with it in any way.

neoKushan:
I've just done some tests and unfortunately, the "private DNS" settings of Android P seem incompatible with Blokada (and VPN's in general). 

Even when Blokada is set to use the "default DNS" of the device, it reverts back to the default DNS given by DHCP rather than the private DNS the user specified.

When cloudflare is explicitly selected in Blokada, it uses the standard DNS lookup instead of DNS-over-TLS (hardly unexpected).

I think DNS-over-TLS needs to be baked into Blokada itself.

iron2000:
Don't know about Android coding but there mention of an API in an Android Dev blog post:
https://android-developers.googleblog.com/2018/04/dns-over-tls-support-in-android-p.html

Maybe it is possible to send the dns request to Pie's Private DNS through this API?
Make a new DNS option in the Blokada list to use Pie's Private DNS?

jeroen7s:
@iron2000 lookignat the api, seems like the only thing it does is returning true or false ibased on wether android's current DNS is encryptede or nor, don't see anything related to changing/setting up a secure DNS connection

Dr-Flay:
DNSCrypt may be the way to add not only encrypted DNS, but to also include DNS validation.
All the resolvers are OpenDNS, but unlike the official Cisco ones, many do not log. 32 currently with DNSSec including Cloudflare
[...]

PR #558, opened 2019-11-01
Implement DNS over TLS

tkterris:
Implement DNS over TLS functionality. Add a field "dotEnabled" to the DNS configuration model (Dns and DnsChoice), and a step in the custom DNS creation logic to support adding the DNS server hostname (e.g. 1dot1dot1dot1.cloudflare-dns.com). 

This resolves issue https://github.com/blokadaorg/blokada/issues/198


tkterris:
Too much of a performance impact, declining the PR.

tkterris:
Still experiencing performance issues. Closing again.

tkterris:
Turns out the "performance issues" were caused by not tying the TCP connection into the main loop properly. That's been resolved.

app/src/ui-blokada/kotlin/core/Dns.kt
```diff
@@ -191,23 +202,36 @@ class DnsChoicePersistence(xx: Environment) : PersistenceWithSerialiser<List<Dns
 
 }
 
-private fun addressToIpString(it: InetSocketAddress) =
-        it.hostString + ( if (it.port != 53) ":" + it.port.toString() else "" )
+private fun addressToIpString(it: InetSocketAddress, dotEnabled: Boolean) =
```
tkterris: The encoding and decoding (`addressToIpString`, `ipStringToAddress`, and `ipStringToDotEnabled`) is pretty terrible. To support backwards compatibility, the DoT hostname is appended to the IP address, with "#" as the delimiter. So:
```
8.8.8.8#dns.google;8.8.4.4#dns.google
```
This is obviously not great, any suggestions here would be appreciated.

app/src/ui-blokada/kotlin/core/AddDnsActivity.kt
```diff
@@ -86,9 +88,32 @@ class AddDnsActivity : Activity() {
...
+                var portString = "853"
+                if (it.contains(":")) {
+                    val parts = it.split(":")
+                    host = parts[0]
+                    portString = parts[1]
+                }
+                var port = portString.toInt()
+                for (i in servers.indices) {
+                    if (servers[i] != null) {
+                        servers[i] = InetSocketAddress(InetAddress.getByAddress(host, servers[i]!!.address.address), port)
```
tkterris: This logic could probably be better. If the user specifies a DoT hostname, we iterate over the previously specified servers, changing them to InetSocketAddresses with the DoT hostname as the hostname and the previously entered IP address as the IP address. We then set the `dotEnabled` flag. It's a bit awkward.

app/src/full/kotlin/tunnel/DnsProxy.kt
```diff
@@ -120,6 +129,34 @@ internal class DnsProxy(
...
+
+            DataOutputStream(s.outputStream).use {
+                //send TCP request
+                it.writeShort(rawData.size)
+                it.write(rawData)
+                it.flush()
+            }
+
+            if (originEnvelope != null) forwarder.add(s, originEnvelope)
+            else Result.of { s.close() }
```
ps100000: instead of closing the socket if the envelope is `null` it would be better if the socket wouldn't even be created.

app/src/ui-blokada/kotlin/core/AddDnsActivity.kt
ps100000: The best option might be to rewrite the first input field to also allow a valid hostname instead of a IPv4 address and to change the rest of the steps accordingly. Validation of the hostname would be nice either way,

tkterris: Added hostname validation. 

I don't think having a hostname instead of an IP address makes sense, though. The IP address is what's used for the DNS lookup, the hostname is just for verifying the cert. See this test, where both the IP address and hostname are specified: https://getdnsapi.net/query/

app/src/full/kotlin/tunnel/DnsProxy.kt
tkterris: I could make that change, but that would differ from how it's being done for the UDP request. Plus, currently it will send the request, just not add it to the forwarder if the envelope is null.

app/src/ui-blokada/kotlin/core/Dns.kt
tkterris: ...and I just realized that, if the DNS list at [here](https://blokada.org/api/v4/content/en/dns.txt) is updated to include the DoT hostnames, it will break for anyone using an earlier version of Blokada (since it won't expect the hostname to be appended to the IP address). The best solution would probably be to create a new dns.txt with a separate field for the TLS hostname, and use that.

ps100000: I'm currently rewriting the API to use JSON files instead of the current format so this won't be a problem once I'm finished and the next API version is used.

app/src/ui-blokada/kotlin/core/AddDnsActivity.kt
ps100000: I should have been a bit more clear. There would be 2 possibilities:
User enters a IP in the first field-> User is prompted to input another IP for normal DNS 
User enters a domain in the first field-> User is prompted to input the IP

But I don't know if this would be confusing,

tkterris: How about adding a new button to the DNS page, in addition to the "Add new DNS..." button, "Add new DNS-over-TLS..."? If the user selects the DNS button, the flow is the same as before these changes:
```
IP 1 -> IP 2 -> Display Name
```
and if they select the DoT button, the flow is:
```
IP 1 -> IP 2 -> DoT Hostname -> Display Name
```
For each flow, all fields would be required.

tkterris:
Should there be some indication in the UI that a DNS server is using DoT? Perhaps by having a small lock icon next to the display name in the DNS server list? As it is, once the DNS server is selected, there's no indication that Blokada is using DoT. I've just been visiting [1.1.1.1/help](https://1.1.1.1/help) to confirm that DoT is being used. Thoughts?

app/src/full/kotlin/tunnel/DnsProxy.kt
tkterris: Nevermind, I misunderstood. The nullcheck is only used for empty UDP packets, which doesn't occur here. I've removed the nullcheck.

app/src/ui-blokada/kotlin/core/AddDnsActivity.kt
tkterris: Looks like adding a branching flow would involve rewriting VBStepView, since currently it only supports a single list of ViewBinders. The simplest options would be to have an optional DoT hostname field (like is in the PR now), or to have two different flows, one for standard DNS and one for DoT. The DoT flow can be simplified to:
```
DoT Hostname -> Display Name
```
by using getAllByName, so that could simplify things if it has its own flow.

tkterris: Per above comment, created a separate flow, with a new button, for creating a DNS over TLS server. The change is in commit 0eeaa9a.

ps100000: I think this is the best solution. :+1:

ps100000:
> Should there be some indication in the UI that a DNS server is using DoT? Perhaps by having a small lock icon next to the display name in the DNS server list? As it is, once the DNS server is selected, there's no indication that Blokada is using DoT. I've just been visiting [1.1.1.1/help](https://1.1.1.1/help) to confirm that DoT is being used. Thoughts?
You could change the server-like icon thats used for all DNS-related settings and change it to a lock. Maybe in a different color.

tkterris:
> > Should there be some indication in the UI that a DNS server is using DoT? Perhaps by having a small lock icon next to the display name in the DNS server list? As it is, once the DNS server is selected, there's no indication that Blokada is using DoT. I've just been visiting [1.1.1.1/help](https://1.1.1.1/help) to confirm that DoT is being used. Thoughts?
> 
> You could change the server-like icon thats used for all DNS-related settings and change it to a lock. Maybe in a different/inverted color.

Done. I left the existing server icon for most cases, but when a DoT server is selected (or listed in the DNS selection menu), it reuses the key-inside-a-shield icon. I put in a TODO, in case a new icon should be created.

I noticed that the DoT server will only be used if adblocking and custom DNS are enabled, and the VPN is disabled. This is because the changes were only made to DnsProxy. When the Blokada VPN or DNS without adblocking are used, DnsProxy doesn't intercept the connections, so conventional DNS is used (see TunnelManagerFactory). Resolving these would be architectural decisions outside the scope of this PR (the design of the Google Play Store app for non-adblocking, and the design of the Blokada VPN for connecting to the VPN).

app/src/ui-blokada/kotlin/core/Dns.kt
ps100000: As soon as this got approval you can adapt your PR. :)  https://github.com/blokadaorg/blokada/pull/575

tkterris:
Is there anything else I should do for this PR?

ps100000:
> Is there anything else I should do for this PR?
I should finish the changes @kar requested for API v5 this week.
It would be nice if you could change your code to use the new json-based API after it's merged so we don't have breaking changes twice. That's also the reason why this is not merged by now. :)

tkterris:
> > Is there anything else I should do for this PR?
> > I should finish the changes @kar requested for API v5 this week.
> > It would be nice if you could change your code to use the new json-based API after it's merged so we don't have breaking changes twice. That's also the reason why this is not merged by now. :)

Whoops, I had missed that. I've merged in that branch.

szaimen:
Any progress here? :)

afonari:
https://github.com/IngoZenz/personaldnsfilter has it implemented.

szaimen:
> https://github.com/IngoZenz/personaldnsfilter has it implemented.

could that be used for blokada?

tkterris:
> Any progress here? :)

This change includes the v5 API changes, in this PR:

https://github.com/blokadaorg/blokada/pull/575

Once that's merged in, this will be ready to go.

tkterris:
Obsolete, DOH has been implemented in the latest version of Blokada.

androidacy:
Not really. For instance my college's wifi blocks DoH but not DoT. They're not the same protocol at all

PR #575, opened 2019-11-15
Api v5

ps100000:
Rewrite of the API for filters, dns servers and translations. It replaces the custom file format currently in use with JSON. If you want to test it checkout the commit before the removal of the testdata.


kar:
Do I need sample v5 api data to run it?

kar:
Insead of org.json apis, this should be coded with GSON which is already our dependency for the tunnel api part. This way we can define clear models etc. Check BlockaRest.kt for reference.

kar:
The old FilterSerialiser was also used for loading persisted filters from local storage (to load legacy persistence basically). Did it got broken with this change?

ps100000:
> The old FilterSerialiser was also used for loading persisted filters from local storage (to load legacy persistence basically). Did it got broken with this change?

The old version changed its name to LegacyFilterSerialiser and still exists for that reason.

ps100000:
> Do I need sample v5 api data to run it?

checkout the commit before the last one to get test data. :)

kar:
I noticed the .properties files have changed format too. So now .properties are not java properties format (eg key value), but a json format. That's confusing for one, but also, this will make all translators have to translate all strings in properties files again, plus requires additional work in the crowdin integration and manage5/4 scripts. I think the value of making those files a json is minuscule, better to revert it back to .properties, and to make sure the app expects them to be properties files. The rest seems ok.

ps100000:
I changed back the .properties files. Could you check again? :)

PR #577, opened 2019-11-17
fixed smartlist not working if settings location is internal storage

ps100000:
There was a bug that lead to smart list throwing a exception on internal storage because it tried to safe the files in the root directory. This fixes it and makes the path for persistence etc. less prone to error.


kar:
merging this instead: https://github.com/blokadaorg/blokada/pull/578

PR #589, opened 2019-11-26
Smart list V2

ps100000:
New features: 

- [x] reset button to delete current state
- [x] list-wise loading to get over the memory limit
- [x] cleanup button to check current list for old rules that aren't in any list anymore

Additional:
- refactored config presistence


kar:
I found a bug:
1. clean install
2. activate smart list
3. tap clean (nothing to clean)
4. after some time my rules counter is 0. after some more time it become 1.

PR #590, opened 2019-11-29
Providing memory limitation per app

peterroth:
Applications can't utilise all the available RAM of the device, the Android system limits it. This value is the maxMemory. It is now showed on the UI to provide better explanation why users can't load all the lists.

### DIFF
--- a/app/src/ui-blokada/kotlin/core/bits/Slots.kt
+++ b/app/src/ui-blokada/kotlin/core/bits/Slots.kt
@@ -81,7 +81,7 @@ class FiltersStatusVB(
             content = Slot.Content(
                     label = i18n.getString(R.string.panel_ruleset_title, Format.counter(rules)),
                     header = i18n.getString(R.string.panel_ruleset),
-                    description = i18n.getString(R.string.panel_ruleset_built,
+                    description = i18n.getString(R.string.panel_ruleset_canfit,
                             Format.counter(rules), Format.counter(getMaxMemory()), Format.counter(memory, round = true))
             )
             date = refreshDate
--- a/app/src/ui-blokada/kotlin/core/bits/Slots.kt
+++ b/app/src/ui-blokada/kotlin/core/bits/Slots.kt
@@ -38,6 +38,7 @@ class FiltersStatusVB(
 
     private var rules: Int = 0
     private var memory: Int = 0
+    private var maxMemory = getMaxMemory()
 
     private val updatingFilters = {
         view?.apply {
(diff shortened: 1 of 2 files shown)


app/src/ui-blokada/kotlin/core/bits/Slots.kt
```diff
@@ -38,6 +38,7 @@ class FiltersStatusVB(
 
     private var rules: Int = 0
     private var memory: Int = 0
+    private var maxMemory = getMaxMemory()
```
ps100000: This variable is not needed. Just move the function call to where you use the var. :)

peterroth: Thanks for the feedback.
The variable is removed and the function is called in line 85, changes are committed and pushed.

app/src/ui-blokada/kotlin/core/bits/Slots.kt
```diff
@@ -68,14 +68,21 @@ class FiltersStatusVB(
         Unit
     }
 
+    fun getMaxMemory(): Int {
+        val runtime = Runtime.getRuntime()
+        val totalmem = runtime.maxMemory()
+        val mb = totalmem.div(1048576)
```
kar: What is this magic number?

app/src/ui-blokada/res/values/strings_panel.xml
```diff
@@ -37,7 +37,7 @@
     <!-- How many rules are configured. Parameters is number of rules. -->
     <string name="panel_ruleset_title">Filters now have %s rules.</string>
     <string name="panel_ruleset_updating">Updating filters...</string>
-    <string name="panel_ruleset_built">Filters are now configured with &lt;b&gt;%s&lt;/b&gt; rules. Up to &lt;b&gt;%s&lt;/b&gt; rules can fit into your device memory overall. Remember, this is just an estimate and subject to change depending on other apps on your device, so you should stay well below this limit to be safe.</string>
+    <string name="panel_ruleset_built">Filters are now configured with &lt;b&gt;%s&lt;/b&gt; rules. Your device allows max &lt;b&gt;%s&lt;/b&gt; MB of memory for every application, and up to &lt;b&gt;%s&lt;/b&gt; rules can fit into this allocated value.&lt;br&gt;Remember, this is just an estimate and Blokada utilises the most memory when it merges the lists.</string>
```
kar: When you change the number of placeholders in a string, please introduce a new string (change its name), or there might be problems with old translations.

kar: Maybe change "allocated value" with "it", this sounds too scary

app/src/ui-blokada/kotlin/core/bits/Slots.kt
peterroth: The return value is bytes, to display it in MB, I have to divide the result 2 times with 1024. That's the magic number.

app/src/ui-blokada/res/values/strings_panel.xml
peterroth: Ok, I'll keep it in mind. Fixed in a new commit.

PR #595, opened 2019-12-06
Changing memory related values

peterroth:
Based on a list with 1.327.728 unique hostnames, the average hostname lenght is 20,41 char long. As fractions aren't supported, it's better to round up to 21.
Unicode characters use 2 bytes.
I ran a test for a few days to see how much RAM one hostname consumes in average, it is around 130 bytes

### DIFF
--- a/app/src/tun-local/kotlin/tunnel/Model.kt
+++ b/app/src/tun-local/kotlin/tunnel/Model.kt
@@ -53,7 +53,7 @@ class Memory {
         val linesAvailable = {
             val r = Runtime.getRuntime()
             val free = r.maxMemory() - (r.totalMemory() - r.freeMemory())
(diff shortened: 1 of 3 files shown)


peterroth:
4 person tested it so far and no crashes were reported.

kar:
Could you sum up the beta tests results so I can merge then?

peterroth:
12 person tested the version with increased memory limit with various devices, some of them older one. No one reported app crashes, most of the users were able to load all host lists, one after another.
After the system triggered a GC, they were able to enable additional lists without issues or warning. 
Take away of the test: if multiple lists are enabled at the same time, Blokada shows all of them enabled while merging, but later, when the merge is done and the UI refreshes some of them are disabled. I think this is the default behavior, not caused by the changes.

Bottom line: the app has a limited memory, enabling too much lists at the same time will fill that up while merging them, but after a GC it's possible to add more. No crash or restart was reported.

Issue #495, opened 2019-08-28
I want to be able to turn off "Host Log"


nu111:
I would like to be able to turn off "Host Log", or maybe to password protect (with fingerprint support) the access to Blokada app.

PrintableCharacter:
Hi,
You can always use third party apps to protect blokada with a password or your fingerprint :)
Sure. That's not the same. But it's a workaround ^^

PR #614, opened 2020-01-05
Update logic to block request logging

tkterris:
Update the "Disabled" request logging option to also block logging in memory / local storage. Added a new option, "Log to internal storage only", which is selected by default, and matches the current functionality.

Resolves issue #495 .

### DIFF
--- a/app/src/ui-blokada/kotlin/core/Tunnel.kt
+++ b/app/src/ui-blokada/kotlin/core/Tunnel.kt
@@ -298,9 +298,14 @@ fun newTunnelModule(ctx: Context): Module {
                 }
             }
 
-            ktx.on(TunnelEvents.REQUEST_SAVED) { request ->
+            ktx.on(TunnelEvents.REQUEST) { request ->
                 if (request.blocked) {
                     s.tunnelDropCount %= s.tunnelDropCount() + 1
+                }
+            }
+
+            ktx.on(TunnelEvents.REQUEST_SAVED) { request ->
+                if (request.blocked) {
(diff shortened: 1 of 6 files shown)


app/src/ui-blokada/res/values/strings_logger.xml
```diff
@@ -4,6 +4,7 @@
     <string name="logger_slot_title">Request logger</string>
     <string name="logger_slot_desc">When enabled, Blokada will create a CSV file and log requests that have been allowed and / or blocked by the app. Look for "blokada" folder on your external storage.</string>
```
ps100000: Description doesn't match the new functionality.

app/src/ui-blokada/kotlin/core/Tunnel.kt
```diff
@@ -298,7 +298,7 @@ fun newTunnelModule(ctx: Context): Module {
                 }
             }
 
-            ktx.on(TunnelEvents.REQUEST) { request ->
+            ktx.on(TunnelEvents.REQUEST_SAVED) { request ->
```
ps100000: The counter for the number of blocked requests will break with this change

app/src/ui-blokada/res/values/strings_logger.xml
tkterris: Done.

app/src/ui-blokada/kotlin/core/Tunnel.kt
tkterris: Fixed.

kar:
Thank you @tkterris for your contribution!

PR #674, opened 2020-04-04
BUILD add and Bash edits

s4pun1s7:
* BUILD.md still needs some work and possibly additional information

* The two `.rs` files changed are because of  `rustc` warning of deprecation.

* `install.sh` script made to directly call `boringtun/build.sh` to avoid errors regarding paths

* `build.sh` added an export to set the ndk bundle path to `~/Android/android-ndk-r21` (latest ndk as of yet)
* * also added sed commands to update the `boringtun/.cargo/config` to linux

I've run multiple test builds and managed to installFullDebug to my android device
Latest successful build with:
```
cargo 1.42.0 (86334295e 2020-01-31)
rustc 1.42.0 (b8cedc004 2020-03-09)
```

### DIFF
--- a/boringtun/build.sh
+++ b/boringtun/build.sh
@@ -1,7 +1,13 @@
 #!/bin/sh
 JNI_LIBS=../app/src/tun-blocka/jniLibs
 
-export PATH=$PATH:$NDK_STANDALONE/toolchains/llvm/prebuilt/linux-x86_64/bin
+# Change the next export as needed. It should point to your android ndk bundle
+export NDK=$HOME/Android/android-ndk-r21
+export PATH=$PATH:$NDK/toolchains/llvm/prebuilt/linux-x86_64/bin
+
+# The next two sed commands change some paths...
+sed -i -e "s|/Users/kar/Library/Android/sdk/ndk-bundle|$NDK|" .cargo/config
+sed -i -e "s|darwin|linux|" .cargo/config
 
 rm -rf $JNI_LIBS
 mkdir $JNI_LIBS
@@ -28,6 +34,7 @@ export CXX=i686-linux-android21-clang++
 cargo build --lib --release --target i686-linux-android
 cp target/i686-linux-android/release/libboringtun.so $JNI_LIBS/x86/libboringtun.so
 
+# Comment this out to save some time
 echo "Building for v86_64..."
 export CC=x86_64-linux-android21-clang
 export CXX=x86_64-linux-android21-clang++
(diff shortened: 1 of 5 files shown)


boringtun/src/ffi/benchmark.rs
```diff
@@ -49,7 +49,7 @@ fn format_float(number: f64) -> String {
     formatted
 }
 
-fn run_bench(test_func: &mut FnMut() -> usize) -> f64 {
+fn run_bench(test_func: &mut dyn FnMut() -> usize) -> f64 {
```
kar, line 52: Why did you have to change this? We do not modify the boringtun directly, we fork it and do git subtree. Fork is here: https://github.com/kar/boringtun

boringtun/src/noise/mod.rs
```diff
@@ -64,7 +64,7 @@ impl FromStr for Verbosity {
     }
 }
 
-type LogFunction = Box<Fn(&str) + Send>;
+type LogFunction = Box< dyn Fn(&str) + Send>;
```
kar, line 67: same here

boringtun/build.sh
```diff
@@ -1,7 +1,13 @@
 #!/bin/sh
 JNI_LIBS=../app/src/tun-blocka/jniLibs
 
-export PATH=$PATH:$NDK_STANDALONE/toolchains/llvm/prebuilt/linux-x86_64/bin
+# Change the next export as needed. It should point to your android ndk bundle
+export NDK=$HOME/Android/android-ndk-r21
```
kar, line 5: if you are fixing hardcoded paths maybe we should not assume that someone has $HOME/Android

install.sh
```diff
@@ -1,31 +1,3 @@
...
-
-echo "Building for i686..."
-export CC=i686-linux-android-gcc
-export CXX=i686-linux-android-g++
-cargo build --lib --release --target i686-linux-android
-cp target/i686-linux-android/release/libboringtun.so $JNI_LIBS/x86/libboringtun.so
-
-cd ../
-./gradlew iFD
+cd boringtun; . ./build.sh; cd ..; ./gradlew iFD
```
kar, line 3: We will have to make sure that fdroid build does not break. We'll have to modify the build configuration here: https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/org.blokada.alarm.yml

boringtun/src/ffi/benchmark.rs
s4pun1s7: sorry about that. because the rustc warning I changed it to see if it'll work and it did. Noted

PR #718, opened 2020-05-27
notification: Take at most 15 blocked history items

MarijnS95:
This fixes a nasty crash when toIndex in subList goes out of range:
(Cut down for brevity)
```
E AndroidRuntime: java.lang.RuntimeException: Unable to bind to service core.KeepAliveService@8ecb0ab with Intent { act=KeepAliveService cmp=org.blokada.origin.alarm/core.KeepAliveService }: java.lang.IndexOutOfBoundsException: toIndex = 15
[...]
E AndroidRuntime: 	at java.util.ArrayList.subListRangeCheck(ArrayList.java:1016)
E AndroidRuntime: 	at java.util.ArrayList.subList(ArrayList.java:1008)
E AndroidRuntime: 	at notification.UsefulKeepAliveNotification$1.invoke(ANotificationUtils.kt:184)
E AndroidRuntime: 	at notification.UsefulKeepAliveNotification$1.invoke(ANotificationUtils.kt:160)
E AndroidRuntime: 	at notification.BlokadaNotificationManager.getNotification(ANotificationUtils.kt:71)
```

4671acacb484a0151ec0046e3289763f6f8de7c2 applies a filter over history
which has a tendency to return fewer items than `logSublistEnd`, resulting
in the above bounds check failure. `.take` has no such limitations: it'll
give at most 15 of the desired items.

---
Note that I don't fully understand the:
```kt
} else if (logSublistEnd > 0) {
    logSublistEnd--
}
```
clause at the end. `toIndex` in `subList()` is exclusive according to [the docs](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list/sub-list.html) meaning a list with one item results in `.subList(0, 0)` and hence no history elements as a result. Is this by accident or should this be replicated?

### DIFF
--- a/app/src/ui-blokada/kotlin/notification/ANotificationUtils.kt
+++ b/app/src/ui-blokada/kotlin/notification/ANotificationUtils.kt
@@ -175,13 +175,7 @@ class UsefulKeepAliveNotification(val count: Int, val last: String): BlokadaNoti
             } else {
                 val domainList = NotificationCompat.InboxStyle()
 
-                var logSublistEnd = RequestLog.getRecentHistory().size
-                if(logSublistEnd > 15) {
-                    logSublistEnd = 15
-                } else if (logSublistEnd > 0) {
-                    logSublistEnd--
-                }
-                RequestLog.getRecentHistory().filter { it.state == RequestState.BLOCKED_NORMAL }.subList(0,logSublistEnd).asReversed().distinct().forEach { request ->
+                RequestLog.getRecentHistory().filter { it.state == RequestState.BLOCKED_NORMAL }.take(15).asReversed().distinct().forEach { request ->
                     domainList.addLine(request.domain)
                 }
 


MarijnS95:
If only there was an option to fork a repository without taking all branches and tags with it I wouldn't accidentally remove the PR'ed branch all the time :grimacing:

ps100000:
Looks like it was 3 AM when I wrote this. ^^
Thx for fixing it and pointing out a better solution with take().
Didn't know that function exists. :)
There was another place which used (almost) the same code.
Changed it in #720 .

MarijnS95:
Yeah it's a nifty little cleanup of the code, and prevents this exception at the same time. I wasn't aware of `.take()` either but used it in other languages and wasn't surprised to see it in Kotlin too :)

kar:
Good job guys!

kar:
@MarijnS95 seems like a crash still happens in prod:

```
06-07 16:10:42.781 10464 10464 E AndroidRuntime: Caused by: java.util.NoSuchElementException: Collection contains no element matching the predicate.
06-07 16:10:42.781 10464 10464 E AndroidRuntime: 	at notification.UsefulKeepAliveNotification$1.invoke(ANotificationUtils.kt:295)
06-07 16:10:42.781 10464 10464 E AndroidRuntime: 	at notification.UsefulKeepAliveNotification$1.invoke(ANotificationUtils.kt:160)
06-07 16:10:42.781 10464 10464 E AndroidRuntime: 	at notification.BlokadaNotificationManager.getNotification(ANotificationUtils.kt:71)
06-07 16:10:42.781 10464 10464 E AndroidRuntime: 	at notification.BlokadaNotificationMain$getNotification$1.invokeSuspend(ANotificationUtils.kt:43)
06-07 16:10:42.781 10464 10464 E AndroidRuntime: 	at notification.BlokadaNotificationMain$getNotification$1.invoke(Unknown Source:10)
```

Are you sure take() takes at most n items, even if collection has less (or zero)?

MarijnS95:
@kar Yes I am 100% sure, see the following Kotlin playground example: https://pl.kotl.in/-smvDNFyI

What's the version of this build? Line `295` does not even exist in the master version of ANotificationUtils.kt: https://github.com/blokadaorg/blokada/blob/master/app/src/ui-blokada/kotlin/notification/ANotificationUtils.kt

kar:
It's supposed to be 4.6.4 but I agree line numbers don't match up for some reason. I'm just trying to eyeball where that predicate crash could happen...

MarijnS95:
A bit of introspection shows that this is the likely cause:

https://github.com/blokadaorg/blokada/blob/9769091e76af91aac125dc8204833bab663d4269/app/src/ui-blokada/kotlin/notification/ANotificationUtils.kt#L164

kar:
That must be it. Nice one, thanks a ton!

PR #720, opened 2020-05-27
refactored code in ListWidget to use take() 

ps100000:


MarijnS95:
I doubt this runs out of bounds as the filtering is applied after the `subList`, unless the list is shortened asynchronously between calculating `logSublistEnd` and running the `subList` function (which may not be touched async at all, I haven't looked at the rest of the program).

Rather, should the `.filter` and `.take` be swapped around? In the current instance if the last 50 history items all turn out not to be blocked by the blacklist no item is shown.

ps100000:
It was meant more as a cleanup but sure it wont hurt to swap the 2. Though I don't think that it really matters that much.

MarijnS95:
@ps100000 I assume most items are `BLOCKED_NORMAL` indeed, but you never know how many exceptions are set up etc, better be safe than sorry :)

MarijnS95:
It anway dosn't matter since the `widget_list_message` `TextView` is limited to `android:maxLines="10"`, anything beyond that is ellipsized even if the widget is resized to be longer. If no scrolling is desired how about removing `maxLines` and instead `.take()` a number that fits the `TextView` in the current layout?

kar:
If you find a way let us know.

PR #738, opened 2020-06-18
change RequestLog to match based on request id instead of domain

ps100000:
and fixed a possible out of bounds exception.


kar:
Could you explain why do we want to match on request id?

ps100000:
The advantage is that request ids are (unique) random numbers. So it's much more accurate to use them instead of the domain. :)

ps100000:
It is used for every (future) function that has to update an entry in the log based on what the answer of the DNS server was. for example currently the RCODE of the answer is checked to tell if the DNS resolved the domain.

PR #753, opened 2020-07-17
changed counter on dashboard to also display start date

### LINKED ISSUE #594: I want to see since when the counter is running
I want to see since what date the counter on the main page is running.
xx.xxx Blocked hosts since dd.mm.yyyy

ps100000:
Fixes #594

### DIFF
--- a/app/src/ui-blokada/res/values/strings_bits.xml
+++ b/app/src/ui-blokada/res/values/strings_bits.xml
@@ -14,7 +14,7 @@
     <string name="home_dns_only">Blokada is active in DNS mode</string>
 
     <!-- A text on the main ads counter. "490 requests blocked" -->
-    <string name="home_requests_blocked">%s blocked since %s</string>
+    <string name="home_requests_blocked_since">%s blocked since %s</string>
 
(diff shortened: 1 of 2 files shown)


app/src/ui-blokada/res/values/strings_bits.xml
```diff
@@ -14,7 +14,7 @@
     <string name="home_dns_only">Blokada is active in DNS mode</string>
 
     <!-- A text on the main ads counter. "490 requests blocked" -->
-    <string name="home_requests_blocked">%s blocked</string>
+    <string name="home_requests_blocked">%s blocked since %s</string>
```
kar, line 17: Please introduce a new string for that. Changing the number of placeholders in a preexisting string, messes up the existing translations and afair leads to crashes

ps100000: Is it sufficient to rename it or does the old string still has to exist?

kar: Old string does not have to exist.

PR #769, opened 2020-08-05
Added CNAME-Blocking and refined detection of requests blocked by the DNS server

ps100000:


kar:
Quite nice, thanks!

This feature has grown a bit, and now people don't understand what kind of events should they see in the hosts log. I think it would be a good idea to write a little documentation on the new cname blocking and the answer types users can expect to see and what do they mean.

app/src/full/kotlin/tunnel/BlockaTunnelFiltering.kt
```diff
@@ -148,20 +149,49 @@ internal class BlockaTunnelFiltering(
 
     private fun rewriteSrcDns4(packet: ByteBuffer, length: Int) {
         val originEnvelope = try {
-            IpSelector.newPacket(packet.array(), packet.arrayOffset(), length) as IpPacket
+            IpSelector.newPacket(packet.array(), packet.arrayOffset(), length) as IpV4Packet
```
kar, line 152: What about ipv6?

Issue #798, opened 2020-11-24
Blokada counts allowed hosts as blocked


peterroth:
The block counter on Blokada's main screen counts not only blocked domains, but allowed ones too. If I open [blokada.org](https://blokada.org), the counter increases by 10, however, nothing is being blocked (based on activity screen).

App version: Blokada 5.5
Device model: OnePlus 7 pro
OS version: Oxygen OS Open Beta 19
Steps to reproduce the bug: open Blokada, note the blocked host counter. Then open a website that is surely allowed and check the counter again.
Expected behavior: allowed hosts shouldn't be reported as blocked
Actual behavior: allowed hosts are counted like they are being blocked

peterroth:
Issue is fixed in 5.6 (currently in beta, stable will be released this Friday).

PR #802, opened 2020-12-04
Fixing issue #798

peterroth:
The first commit changes the getStat's denied variable, the other fixes the counter on the main screen to report the correct value of blocked elements (instead of the allowed ones).


peterroth:
After a discussion with @kar, the getStat was correct as it was, setting that to runtimeDenied duplicates the entries.

PR #846, opened 2021-03-10
Update Alternate's IPs and add DoH

peterroth:
Based on Alternate's website, they have new IP addresses and extended their service with DoH: https://alternate-dns.com/


kar:
Can't support binary DoH yet. But updated the plaintext IPs

PR #851, opened 2021-03-17
Bypass Threema HMS builds as well

dbrgn:
Threema recently published their app in the Huawei AppGallery. Those
builds have a different package name compared to Google Play:

- ch.threema.app.hms
- ch.threema.app.work.hms

Bypass those as well.


dbrgn:
Source: https://twitter.com/ThreemaApp/status/1372203872970756102

Thank you!

PR #859, opened 2021-04-22
Correction for Description and Slugline of CPBL

bongochong:
Just correcting the description and slugline for my list in the iOS version.


bongochong:
> Why did you open this twice?

One for the iOS version, and one for the Android version. Two different files.

androidacy:
Why can't you do both in the same pr? Fork it then clone it locally and push changes to your fork, then open a single pr

bongochong:
Didn't think about it that way, but done and done. Sorry about that.

Issue #833, opened 2021-02-11
Blokada 5 - no landscape mode?


rob-X1:
I just installed Blokada 5 and was a bit surprised that the UI seems to completely missing a landscape mode. 

On a tablet that is used only in landscape mode this is really annoying.

peterroth:
Correct, Blokada v5 has no landscape mode.

imanincel:
Can we add that as a **[Feature Request]**?

PR #884, opened 2021-08-10
allow MainActivity to be shown in landscape mode

jpstotz:
This change allows users to use Blokada in landscape mode (see #833).

Tested on a 10" tablet.

### DIFF
--- a/android5/app/src/main/AndroidManifest.xml
+++ b/android5/app/src/main/AndroidManifest.xml
@@ -49,7 +49,7 @@
             android:name="ui.MainActivity"
             android:label="@string/app_name"
             android:launchMode="singleTask"
-            android:screenOrientation="portrait">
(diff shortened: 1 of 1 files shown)


kar:
Sorry but I can't merge it, we disabled it on purpose, the UI is not adapted to all screen sizes to work in landscape. This would break small screen devices etc. More work would need to be done. If your change would only apply to tablet screens, then I could consider merging it.

jpstotz:
@kar Would you accept a solution that does allow landscape mode if the screen size is e.g. larger than 9 inch screen in diagonal?

On a device with a smaller display Blokada would start with the "splash screen" (white background with Blokada logo and nothing more)  in landscape mode (if the device is held this way). Then calculate the screen size and switch to portrait mode (device rotates) and finally switch to the "start fragment" (the one with the large power button).

PR #887, opened 2021-08-25
Include more DoH Servers and correct some DNS values

### LINKED ISSUE #836: Quad9 DoH
App version: 5.8.0

When using the "Networks" feature to choose a DNS server, it appears Quad9 is only available through unsecured DNS:

https://github.com/blokadaorg/blokada/blob/8692f22c47e56836e300bccb3e231edcbf3d1a0c/android5/app/src/main/java/repository/DnsDataSource.kt#L140-L144

Even though the service does support DoH: https://www.quad9.net/service/service-addresses-and-features#doh

xopez:
Again i have filled out some DoH servers. Someone has to double check, what I have changed since I am tired a bit.
And I have a question regarding the digitalcourage dns configuration.
Is it really needed to set the same IP two times? Since we speak if `listOf` it should only one be needed.

Fixes #836

### DIFF
--- a/android5/app/src/main/java/repository/DnsDataSource.kt
+++ b/android5/app/src/main/java/repository/DnsDataSource.kt
@@ -160,7 +160,7 @@ object DnsDataSource {
         ),
         Dns.plaintextDns(
             id = "verisign",
-            ips = listOf("64.6.64.6", "64.6.65.6"),
+            ips = listOf("64.6.64.6", "64.6.65.6", "2620:74:1b::1:1", "2620:74:1c::2:2"),
             label = "Verisign Public DNS"
         )
     )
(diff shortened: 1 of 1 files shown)


xopez:
Just saw this: https://github.com/blokadaorg/blokada/issues/836#issuecomment-790626521
I was probably to fast.

kar:
Thank you for the PR. Indeed, we cannot merge it at the moment. We could if you split up the non-doh changes, like updated IPs. Although the trick here is that it'd be necessary to test out those configurations actually work before we merge.. 

Otherwise I'll leave the PR here for a while until we can get around the doh standard.

xopez:
Thinking about the mass of commits in a PR, I'd like to leave it open like that so that we don't get confused there.

Issue #927, opened 2022-03-08
DoH: Support RFC 8484 conform DNS over HTTPS


ryru:
According to https://github.com/blokadaorg/blokada/issues/836 Blokada's DoH support is limited to [Googles JSON-based DoH specification](https://developers.google.com/speed/public-dns/docs/doh/json), which is not an Internet standard. [Most public DoH resolvers](https://github.com/curl/curl/wiki/DNS-over-HTTPS#publicly-available-servers) do not support JSON-based DNS.

Please add DoH support for the official [RFC 8484 DNS Queries over HTTPS (DoH)](https://datatracker.ietf.org/doc/html/rfc8484).

balboah:
Hello,

With Blokada v6 we use the DoH/DoT clients provided natively by the OS, this this issue is only relevant to versions prior to v6.
Of course if anyone would want to try implementing a new client also for v5 we would be happy to welcome such contribution.

PR #928, opened 2022-03-08
Remove non-JSON DoH resolver until RFC 8484 is supported

ryru:
DoH resolver of [Digital Society Switzerland](https://www.digitale-gesellschaft.ch/) don't support DoH based on JSON. Although it was never officially supported, it had worked. Users of Blokada who select these name resolvers are facing name resolution issues.

As soon as #927 is implemented these DoH resolvers can be enabled again.


ryru:
@kar is this project / code repo still alive?

PR #984, opened 2025-03-17
Six translations

balboah:
- small adjustments to the app store description
- add missing arabic translation


kar:
There is a problem with additions:

`13:46:05 /opt/homebrew/Cellar/fastlane/2.226.0_1/libexec/gems/fastlane-2.226.0/spaceship/lib/spaceship/connect_api/api_client.rb:223:in 'Spaceship::ConnectAPI::APIClient#handle_response': An attribute value has invalid characters. - Description can’t contain the following character(s): 🚀. - /data/attributes/description (Spaceship::UnexpectedResponse)
`

I think for ios all emojis have to be dropped afair

PR #987, opened 2025-04-01
metadata: update iOS

balboah:


balboah:
@kar small iterative update for iOS listing

PR #993, opened 2025-05-29
Update android metadata

balboah:


kar:
Changes look ok, but you dropped the trial mention at the beginning, it was intentional to let users this is a paid app and prevent low ratings because they are surprised, I would put it back.

PR #996, opened 2025-06-10
common/payment: sync account attributes with Adapty

balboah:
If there are attributes on the account, also set those on the Adapty profile.

Since Adapty custom attributes are limited, we filter out values such as timestamps as well as convert values to compatible format such as string `"true"` instead of boolean


balboah:
re-formated with android studio, hope it looks a bit better. I'll try to get Zed to format the same way if it does

kar:
Ah yeah now it looks fine. I may have gotten confused because i checked it commit by commit. Formatting seems a bit off in the PaymentBinding.kt but indeed there is no file changed now that has only-formatting changes. It's ok, you don't have to try to make exact same formatting (may be android IDE thing - with flutter it's easier since there is an official dart format tool).

I'll pull it now and see if the android side works OK.

balboah:
ok I've resolved the conflicts

kar:
A small merge mistake with duplicated interfaces. But I was able to run it no problem after fixing, behavior seems nominal, no errors in log. I will merge it and fix in main to not bother you, since I can't push to the PR.

PR #997, opened 2025-06-16
add .editorconfig for shared code style

balboah:
Exported defaults from Android Studio to be shared across IDE's to have consistent code formatting.
Apparently this is supported by VS Code as well as Zed among others


kar:
Perfect

PR #1000, opened 2025-06-24
Blockaweb/expiry

balboah:
Display the different account statuses in web extension popup


kar:
I have to integrate it with my changes before it can get to main, so I have merged it into `freemium` branch and will keep working there until the freemium feature is ready to main.

PR #1001, opened 2025-06-26
pod install

balboah:
Not sure why `pod install` also will change the lockfile, pushing for us to be in sync.
Also re-occuring IOS.xcodeproj change that it really wants to do (xcode 16.4)


kar:
I hope it won't go back and forth. There probably was a reason why I did gitignore this.

PR #1003, opened 2025-07-01
replace content script rules

balboah:
@kar See if you think this this how we can sync blocklist rules into web extension.

I added the blocklists as submodule to not have to duplicate several MB also in this repo, it then uses a convert script to make it become a rules file, which needs to run on build to make sure the file exists.

I'm not sure how the prod building is hooked up with CI, but I added this in both Xcode and ipa-six target.


kar:
Not sure if related to this PR specifically, but I've been using the build in freemium mode, and it seems to me like adblocking is working just fine, but not the youtube video blocking. I could not make it block ads.

I noticed that in system settings the "m.youtube.com" urls were set to "ask", but for my onboarding flow it won't show that dialog popup anymore (as we have shown in the instruction video) - maybe because I keep reinstalling and installing builds, not sure. Anyway I changed it manually from "ask" to "always allow" but it did not make a difference.

PR #1005, opened 2025-07-04
various fixes for issues found in testing

balboah:
The commits aren't really related more than that all is resulting tweaks from running manual testing.
You may want to view each commit diff & description separate to focus on each specific case


kar:
Nice. Code looks good even if I have done it slightly differently (but not really objectively better). Tests added and pass. App runs and seems to behave as wanted. I don't really have anything to complain.

PR #1007, opened 2025-08-29
ios: synchronize HTTP timeouts across all network components

balboah:
- Add HTTP_TIMEOUT_INTERVAL constant (15.0s) in Constants.swift
- Update CommonBinding.swift to use shared timeout constant  
- Fix WgService.swift timeout delays to use HTTP_TIMEOUT_INTERVAL
- Set request timeout in BlockaPacketTunnelProvider.swift
- Add Constants.swift to WireGuard extension targets in project.pbxproj

### DIFF
--- a/ios/App/Assets/Constants.swift
+++ b/ios/App/Assets/Constants.swift
@@ -14,3 +14,6 @@ import Foundation
 
 // Many events initiated from UI are debounced to avoid general abuse (in seconds)
 let DEFAULT_USER_INTERACTION_DEBOUNCE = 0.3
+
+// HTTP timeout configuration - used by main app, network extension, and WgService (in seconds)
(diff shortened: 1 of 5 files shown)


kar:
looks good, I'll run it.

This keeps adding and deleting based on whether built from xcode or from cli. can we do something about it? I think empty fields like that shouldn't be added?

```
			inputPaths = (
			);
			outputPaths = (
			);
			```

balboah:
Those might be related to pods update, not sure. We tried to avoid it previously. 

How do you build & run dev on your phone? In terminal?

kar:
I have 16.4

kar:
I think it's gets "deleted" when running builds from cli. so I don't think it should be included in diffs overall

PR #1042, opened 2026-03-30
cold start stability

balboah:
After the startup refactor that separated background bootstrap from foreground startup and exposed bootstrap identity earlier, a few timing issues became visible:

- the DNS wizard could appear even when DNS was already configured, because the native DNS state was not always available on the first check
- a white screen could appear on slow startup paths, such as when waiting for the account id store to load from iCloud while in airplane mode
- the power button could briefly show the timer layout even when no timer was set

This PR resolves those issues by:

- retrying DNS verification when the native DNS state is temporarily unavailable, and preserving more native DNS state detail instead of collapsing failures into an empty string
- only showing the timer layout when a timer is actually set
- releasing the native splash on the first Flutter frame instead of relying on a fixed delay


common/lib/src/features/private_dns/ui/private_dns_sheet_ios.dart
kar, line 1: Probably analogous changes ar needed for _android version of this file

common/lib/src/platform/app/startup_first_frame_reporter.dart
```diff
@@ -0,0 +1,66 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+class StartupFirstFrameReporter extends StatefulWidget {
```
kar, line 8: No comment why it has to exist

common/lib/src/platform/app/startup_promotion_gate.dart
```diff
@@ -24,42 +22,52 @@ class StartupPromotionGate extends StatefulWidget {
 
 class _StartupPromotionGateState extends State<StartupPromotionGate>
     with WidgetsBindingObserver, Logging {
-  late bool _ready = widget.launchContext.allowRunApp;
+  static const _visibleUiFallbackDelay = Duration(seconds: 1);
+
   bool _promotionStarted = false;
+  bool _promotionCompleted = false;
+  Timer? _visibleUiFallbackTimer;
```
kar, line 36: Why is timer necessary?

kar (changes requested):
Tried to run on Android, will compile but will never settle into loaded state (stuck on "please wait" on start). Log attached

[log-pr.txt](https://github.com/user-attachments/files/26434407/log-pr.txt)

Seems to me that similar changes for privatedns service done for ios, are missing for android side in this PR, maybe that's the problem?

Overall missing the "why" comments for new components. In a month we won't know why something was added. But it's a taste thing, if AI can figure it out I'm fine with not having them in this project.

balboah:
I'll get the android env set up to see what's going on.

regarding comments, perhaps you can suggest a agents.md update with guidelines on what to comment on and when?

balboah:
@codex can you try answering the missing "Why" from the PR questions? brief and to the point please

common/lib/src/platform/app/startup_first_frame_reporter.dart
balboah: Signals the native side that Flutter has rendered its first frame, so the native splash screen can be removed at the right time instead of using a fixed delay.

common/lib/src/platform/app/startup_promotion_gate.dart
balboah: Fallback for when the lifecycle state is already `resumed` before the observer is attached — `didChangeAppLifecycleState` never fires in that case, so without it the app can get stuck on the splash screen.

kar:
> regarding comments, perhaps you can suggest a agents.md update with guidelines on what to comment on and when?

I have added to main now

balboah:
ok ready for another review

kar:
Ok I went through the code as well as built and ran this on android. Did basic onboarding and no-network cold start test.

Behavior seems proper, nice you addressed the modal-dismiss-detection workaround, it was causing some bugs in the past but I couldn't figure out the proper solution before the ai era. I was concerned about the new code not working well but I couldn't bug it out in testing.

I like the new comments. Code changes look reasonable and positive overall. I didn't experience any problems with splash screen / white screen or foreground trigger.

Also built the family flavor, went through onboarding, messed with the lock screen feature and modals (this flavor has more of them), bg/fg switching. All behavior nominal.

lgtm

balboah:
also tested ios fcm event to also confirm it still worked

PR #1044, opened 2026-04-16
ios: add before_paywall essentials onboarding optional flow

kar:
Introduce essentials_onboarding_order account attribute support for iOS freemium
  onboarding and keep the current post-paywall flow as the default. The app now
  parses a typed onboarding-order value from account attributes, routes flagged
  users to the Safari Essentials onboarding before the paywall, and suppresses the
  existing post-paywall onboarding for that override so the flow stays exclusive.

  Also add focused tests covering attribute parsing, default fallback behavior, and
  the guardrail that Essentials is not shown before the paywall unless the API
  explicitly enables it.


kar:
@copilot review this

balboah:
🤖 Verified on iOS 👍

- `before_paywall` → Safari onboarding after power tap, paywall suppressed
- `after_paywall` → paywall opens, no pre-paywall Safari sheet
- attribute absent → defaults to `after_paywall`

Issue #298, opened 2018-12-05
auto update lists


gkalemis:
### Steps to reproduce
1. Start application
2. Change update interval to weekly
3. Close an restart

### Expected behaviour

To start without updating the lists

### Actual behaviour

Always downloads and updates lists

### Configuration
1. Blokada version
Latest 3.6.101401

2. Device model
Xiaomi 5 plus

3. Android version / ROM
8.1.2

### Questions
1. How long have you been using Blokada? 
6 months

2. How long time ago has the problem started to happen? 
When  updated from 3.4

3.  How often does this problem happen to you?
Always

rpeter85:
Blokada downloads the list once, then it activates the schedule you set up. If the first download fails for some reason (download took long, or it still merges the lists) the process will restart the next time.

colbycdev:
It will say updating, but in my experience it doesn't actually update. Mind you, I've noticed that if the last time the app stopped was due to a crash, it'll re-download
________________________________

rpeter85:
Why do you think it doesn't do anything?

colbycdev:
A) it did it once, when I had NO internet
B) file dates
Mind you, I haven't checked every time, but when I did, that was the scenario
________________________________

PR #1049, opened 2026-05-18
v6: weekly privacy report notification touch-up (#298)

balboah:
Closes blokadaorg/issue-tracker#298.

## What changed

- **Body copy** rewritten — old strings were terse and didn't invite the tap. New copy ends with "Tap to see more." on toplist events.
- **Marker–notification invariant** enforced: the FCM handler skips when the event has no renderable in-app marker (`WeeklyReportEvent.isPostable`), and iOS belt-and-braces drops empty payloads at scheduling time.
- **Two notification action buttons**:
  - **"See more"** (primary, foreground) — opens Privacy Pulse, same as a body tap.
  - **"Turn off weekly reports"** (foreground) — opens the app, navigates to Settings, and the Weekly reports toggle visibly animates from ON → OFF for visible confirmation.

## Design notes

The opt-out action was originally a *background* action (iOS) / *broadcast receiver* (Android). Real-device testing surfaced two problems:

1. **iOS hides custom actions on the default banner** — they only surface on long-press / pull-down / Notification Center / lock-screen. Platform constraint, no workaround.
2. **Background-action taps raced the Flutter engine boot.** `CommandsBinding` queues commands in an in-memory array until Flutter calls `doCanAcceptCommands`. If the OS suspends the app before Flutter wakes up, the queue evaporates and the opt-out silently fails. Reproduced empirically.

The final design pivots to foreground actions on both platforms:
- iOS: `UNNotificationAction(options: [.foreground])`.
- Android: `PendingIntent.getActivity(MainActivity::class.java)` with an `id|action` extra, replacing the `BroadcastReceiver`.

Combined with the in-app navigation to Settings, the user gets a clear "I tapped 'turn off' — the app showed me the toggle move, and here's where I can change my mind later." Stronger trust signal for a privacy-branded app than a silent native HTTP PUT, at the cost of an app-foreground.

The toggle flip is timed deterministically: the Settings screen calls `consumePendingOptOutFromNotification` from `addPostFrameCallback`, and the actor delays 700ms internally so the animation lands after the iOS app-launch transition has settled (otherwise the flip completes invisibly during the route slide-in). A 3-second fallback in the actor honors the explicit user intent if Settings never mounts.

## Test plan

- [x] `flutter test test/features/notification/ test/platform/notification/` — 34 tests pass.
- [x] Manually verified on physical iOS device: long-press banner → tap "Turn off weekly reports" → app foregrounds → Settings renders with toggle ON → animates to OFF → backend `PUT /v2/notification/config` lands → re-firing FCM correctly skips the user with `opt_out` reason.
- [ ] Android equivalent verification (not yet retested after the activity-intent pivot — same Dart path runs).
- [x] Code review pass — verdict "with fixes," `UiGen.swift` regen applied, dead `opt out confirmation` key removed.

## Operator follow-ups (after merge)

1. Push the `deps/translate` submodule's `master` branch (currently at `2c5d5a4`) to its remote so Crowdin picks up the new English keys for translation across locales.
2. Consider wiring `swiftgen` into `make translate` as a separate follow-up — stale `UiGen.swift` already drifted twice this PR.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -83,6 +83,14 @@
         <receiver android:name="service.DozeReceiver" />
 
         <receiver android:name="service.NotificationAlarmReceiver" />
+
+        <receiver
+            android:name="service.WeeklyReportActionReceiver"
+            android:exported="false">
+            <intent-filter>
+                <action android:name="org.blokada.weekly.OPT_OUT" />
+            </intent-filter>
+        </receiver>
         <receiver
             android:name="service.BootReceiver"
             android:enabled="true"
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -85,13 +85,6 @@
         <receiver android:name="service.NotificationAlarmReceiver" />
 
         <receiver
-            android:name="service.WeeklyReportActionReceiver"
-            android:exported="false">
-            <intent-filter>
-                <action android:name="org.blokada.weekly.OPT_OUT" />
-            </intent-filter>
-        </receiver>
-        <receiver
             android:name="service.BootReceiver"
             android:enabled="true"
             android:exported="true">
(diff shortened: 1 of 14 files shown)


kar:
lgtm

Issue #160, opened 2018-03-27
I want sharing number of blocked ads on social media


rpeter85:
It would be a great feature if the number of blocked ads could be posted on social media from Blokada's UI, if possible, with the install/first start date of Blokada.

rpeter85:
Share button added in v4 by @ps100000 
Thank you for that!

Issue #71, opened 2017-11-24
Add F-Droid to http://blokada.org/#download


wolftune:
F-Droid includes Blokada, and promoting F-Droid is good for all the values Blokada stands for (software freedom, freedom from ads and tracking etc)

PR #1053, opened 2026-05-19
feat(payment): Android install-referrer → Adapty acquisition-source attribution

kar:
## Context / Why

The Google Ads campaign (issue #71, `blokadaorg/issue-tracker`) ran with **no app-side attribution on Android**, so install→paid events couldn't be tied to the campaign and CPA_paid was unknown. iOS already has this automatically via Adapty's Apple Search Ads integration; **Android had no equivalent for either flavor** (no `updateAttribution`, no Install Referrer; Adapty has no first-class Google Ads integration). This adds the privacy-safe Android counterpart so #71 can resume against a reliable install→paid-by-campaign read.

## SDK-migration decision (the other half of #71)

Evaluated replacing the native Android Adapty SDK with the Flutter SDK: **rejected.** `six` keeps the native SDK solely for in-paywall subscription **proration** (`onAwaitingPurchaseParams`), and the Adapty Flutter Paywall Builder still has no pre-purchase hook (Adapty SDK-Flutter #160 open). Migrating would regress v6. Native SDK + Pigeon bridge kept untouched.

## What changed

- **`play_install_referrer` 0.5.0** (maintained; AGP-8 namespace verified) — captures the Google Play Install Referrer.
- **`install_referrer.dart`** — shared Dart: parse UTM + `gclid`, map to Adapty's `status/channel/campaign/ad_group/ad_set/creative`, stash `gclid` as `acq_*` custom attributes (bypasses the 30-char converter via the existing pre-built-list path). **iOS-gated** so Apple Search Ads is never clobbered (Adapty = one source per profile). Persisted once-per-install idempotency guard.
- **`updateAttribution`** added to the `PaymentChannel` mixin + both impls: Adapty Flutter SDK (`adapty.dart`) for family/iOS; new **Pigeon `doUpdateAttribution`** → native `Adapty.updateAttribution` in `PaymentBinding.kt` for v6 Android (the Flutter SDK is not activated on v6 Android).
- Trigger wired into `PaymentActor.onStart` (post-init, `unawaited`, idempotent). ProGuard keeps for both flavors.

Privacy-safe: no GAID/IDFA/IP (Adapty's collection-disabled flags untouched).

## Testing

- 24 new unit tests (parse / attribution-map / gclid-chunk / idempotency / iOS-gate / transient-error retry); **full suite 228/228 green** incl. untouched `attribute_converter` regression.
- `six` + `family` debug builds **SUCCESSFUL** (Pigeon regen + native Kotlin + plugin compile under AGP 8.13).
- **On-device (OnePlus 7T, Android 12), both flavors:** app starts, attribution path runs and is **accepted by the Adapty backend** (`{"attribution_json":"{...status:organic,channel:google-play...}","source":"custom"}` + `acq_status` custom attr — verified at the HTTP-request level on the native path), idempotency guard sets, **paywalls present** (native `AdaptyPaywallView` for six, `AdaptyUIPaywallPlatformView` for family), no crash. No purchase attempted (not yet testable).
- A multi-agent diff review caught and fixed one blocker (persisted-guard read must use `fetch()` not `now()`), since confirmed on hardware.

## Out-of-band acceptance (not blocking this PR)

Confirm in the Adapty dashboard that a `"custom"` attribution source filters install→paid the same way the integrated Apple Search Ads source does. Real campaign attribution requires a Play internal-testing install (sideloaded builds report organic, as seen on-device).

### DIFF
--- a/android/app/proguard-family.pro
+++ b/android/app/proguard-family.pro
@@ -7,3 +7,6 @@
 -keep class com.adapty.** { *; }
 -keep class com.adapty_flutter.** { *; }
 -keep class com.adapty.flutter.** { *; }
+
+# Google Play Install Referrer (used for Adapty acquisition-source attribution).
+-keep class com.android.installreferrer.** { *; }
--- a/android/app/proguard-rules.pro
+++ b/android/app/proguard-rules.pro
@@ -19,3 +19,6 @@
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
+
+# Google Play Install Referrer (used for Adapty acquisition-source attribution).
+-keep class com.android.installreferrer.** { *; }
--- a/android/app/src/six/kotlin/binding/PaymentBinding.kt
+++ b/android/app/src/six/kotlin/binding/PaymentBinding.kt
@@ -155,6 +155,27 @@ object PaymentBinding : PaymentOps, AdaptyUiEventListener {
         }
     }
 
+    override fun doUpdateAttribution(
+        attribution: Map<String, Any?>,
+        source: String,
+        callback: (Result<Unit>) -> Unit
+    ) {
+        try {
+            Adapty.updateAttribution(attribution, source) { error ->
+                if (error == null) {
+                    log("Sent install attribution to Adapty (native sdk)")
+                    callback(Result.success(Unit))
+                } else {
+                    logError("Failed updateAttribution", error)
+                    callback(Result.failure(error))
+                }
+            }
+        } catch (e: Throwable) {
+            logError("Failed updateAttribution (threw)", e)
+            callback(Result.failure(e))
+        }
+    }
+
     override fun doLogOnboardingStep(
         name: String,
         stepName: String,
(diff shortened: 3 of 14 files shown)


kar:
Both Copilot review comments addressed in 6a51311d:

**1. `SERVICE_UNAVAILABLE` latched as permanent organic** — `fetchReferrer()` now returns a tri-state result. Connection-class codes (`SERVICE_UNAVAILABLE`, `SERVICE_DISCONNECTED`, `DEAD_OBJECT_EXCEPTION`, `BAD_STATE`) are `transientlyUnavailable`: the service sends nothing and leaves the guard unset so it retries on the next app start — a transient Play hiccup can no longer permanently misattribute a paid install as organic. Genuinely permanent codes (`FEATURE_NOT_SUPPORTED`, `DEVELOPER_ERROR`, `PERMISSION_ERROR`, non-Play/sideload) still record organic + latch the guard (no spin).

**2. Guard couples the two Adapty writes** — the guard is now checkpointed immediately after `updateAttribution` succeeds (the acquisition-source signal this PR is for). `setCustomAttributes` (gclid/`acq_*`, BI-only) runs afterwards as best-effort with its own swallow, so a failure there can't unset the guard or cause Adapty's one-source-per-profile `updateAttribution` to be re-attempted/wedged.

Added unit tests for both (transient-unavailable → retry next start; setCustomAttributes-fails-after-updateAttribution-succeeds → guard still set, not re-attempted). Full suite 230 green.

kar:
As discussed, cancelling this work for the time being.

PR #1054, opened 2026-05-19
appium: runner Auto-Lock requirement (real appium-smoke flake fix) + idle-wait hardening

kar:
## Root cause of the `appium-smoke` flakiness: CI-runner device **Auto-Lock**

A controlled experiment settled it. On the **same commit**, with no code change other than the runner iPhone's **Auto-Lock disabled**:

| phase | result |
|---|---|
| device could sleep (Auto-Lock on) | effectively **all FAIL** |
| **Auto-Lock = Never** | **4 / 4 PASS** |

When the iPhone auto-locks/sleeps mid-run, WebDriverAgent returns **black screenshots and an empty/SpringBoard accessibility tree**, so element lookups fail intermittently — surfacing as `element ("~automation.power_toggle") still not existing` or `Failed to find General / Allmänt in Settings`, both passing on re-run. It was **never** the notification-permission race or WDA quiescence (earlier theories in this PR's history were wrong and are reverted).

## The actual fix is runner configuration (documented here, not code)

`automation/appium/README.md` now requires, as one-time device setup: **Settings ▸ Display & Brightness ▸ Auto-Lock ▸ Never**, device kept unlocked and on power. **CI self-hosted-runner devices must have this set.**

## What remains in this PR (low-risk hardening only)

- `appium:waitForIdleTimeout: 0` — legitimate, widely-recommended Flutter+Appium capability; reduces a real class of WDA quiescence stalls regardless of this flake.
- Minimal `acceptNotificationAlert` change (don't bail on the first "no modal dialog" miss; poll within the timeout) + a notification-accept call in `enableDnsProfile`.
- **Reverted**: the speculative aggressive "dismiss SpringBoard prompt without getAlertText" rewrite (unvalidated; not the cause).

Branch history shows the investigation (v1→v2→v3→revert+docs); squash-merge recommended.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,8 +102,12 @@ appium-test:
 	eval "$$(IOS_AUTO_SELECT_FIRST=1 node scripts/setup-session.mjs)"; \
 	export IOS_AUTO_SELECT_FIRST=1; \
 	$(MAKE) -C ../../../ios "$$APPIUM_APP_INSTALL_TARGET" IOS_UDID="$$IOS_UDID" IOS_DEVICE_NAME="$$IOS_DEVICE_NAME"; \
-	xcrun devicectl device process launch --terminate-existing --device "$$IOS_UDID" "$$APP_BUNDLE_ID" >/dev/null \
-		|| echo "wake: device app launch failed (continuing; WDA will activate)" >&2; \
+	echo "appium-test: waking device $$IOS_UDID ($$APP_BUNDLE_ID) before smoke" >&2; \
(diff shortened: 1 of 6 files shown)


kar:
There was two main reasons for flaky smoke test runs:
- system notification prompt timing, would sometimes interfere.
- when device is sleeping (screen off) it seems it can sometimes execute the test, but usually will fail.

Addressed both. the smoke test will wake up and sleep device before/after. tested with 5 runs in a row, all succeeded.

Issue #303, opened 2018-12-13
Blocking npr.org podcasts


Nemi5150:
### Steps to reproduce
1. Install podcast app (I am using https://play.google.com/store/apps/details?id=fm.player)
2. Subscribe to Planet Money NPR podcast 
3. Stream/Download fails with http error message

### Expected behaviour

Podcast should download

### Actual behaviour

Podcast download fails

### Configuration
1. Blokada version
3.6 (101900)
Adguard DNS
Energized BLu and Coinblocker blacklists

2. Device model
Pixel 2 XL

3. Android version / ROM
9

### Questions
1. How long have you been using Blokada? 
9 months

2. How long time ago has the problem started to happen? 
when I updated to 3.6

3.  How often does this problem happen to you?
Every time

rpeter85:
Blokada uses 3rd party lists. It means, it blocks what is on the list.
Try with another list or add the blocked domain or app to whitelist.

Issue #292, opened 2018-11-21
Blokada blocking internet access on OnePlus 6 


w4tadrag:
I keep on noticing that Blokada blocks all internet access on my device from time to time. I believe this issue is very similar to this one https://github.com/blokadaorg/blokada/issues/284. I have to go back to the app and turn it off and on to make my internet work again. Yet I get the issue again. At first I noticed it happening after 12am EST, but I am seeing it happen during the afternoon now. I have done a lot of troubleshooting and have come to the conclusion that it was Blokada.  

### Steps to reproduce
Not sure, as it happens by itself

### Expected behaviour
Shouldn't block access to the internet

### Actual behaviour
Blocks all access to the internet

### Configuration
1. Blokada version: 3.6

2. Device model: OnePlus 6

3. Android version / ROM: 9.0

### Questions
1. How long have you been using Blokada? 
Since ver 2

2. How long time ago has the problem started to happen? 
After installing 3.6

3.  How often does this problem happen to you?
Once or twice per day

colbycdev:
I have an Alcatel(Android 8.1) as well. I know there is a built in function that harms VPN functionality. Since Blokada doesn't run a service, on mine it kills it. On the other Alcatel I had(Android 7.1.1), I had the same issue.
________________________________

colbycdev:
Breaking the built in smart manager helped for me. Though I don't remember when exactly I started using a custom dns...

piehei:
I'm having the same issue. Stock Android (9.0) with OnePlus 6T and latest Blokada.

It randomly stops all network connectivity. This happens multiple times a day. Toggling it fixes the issue.

w4tadrag:
I think I know why it was causing it. I removed it from battery optimization and stopped seeing issues with it for few days now. You guys should try that.

PR #1055, opened 2026-05-20
ios(sim): revive Mocked + FamilyMocked schemes for per-worktree UX iteration

balboah:
## Summary

Makes the iOS `Mocked` and `FamilyMocked` Xcode schemes usable end-to-end on a per-worktree iOS Simulator for fast UX iteration (issue [#303](https://github.com/blokadaorg/issue-tracker/issues/303)). Both flavors land on the proper home screen, hit the real backend (and exercise the iOS↔API wire), without the paywall or first-account-create ritual, and are drivable through the existing Appium harness for view review and full navigation.

The downstream consumer is [#292](https://github.com/blokadaorg/issue-tracker/issues/292) (Family scheduled filtering) — the schedule editor will be iterated against the FamilyMocked sim's device-detail screen reached by this branch.

## Design

Mock at the **system-integration boundary** (what the sim physically cannot run), not at the business-logic boundary (what the sim can run fine). Only Adapty StoreKit2 + the iOS `PrivateDnsService` are forced mocks; Persistence, HTTP, perm bindings, and every other platform channel use their real iOS implementation. The sim talks to `api.blocka.net` / `family.api.blocka.net` normally.

## Load-bearing fix

The Mocked + FamilyMocked targets' `[CP-User] Run Flutter Build common Script` phases now override `FLUTTER_TARGET` after sourcing `flutter_export_environment.sh`. Without this, every Xcode target compiled from `lib/main.dart` (the prod entry) regardless of scheme, and the Dart-side mocked entries were dead code. Production target script phases are unchanged.

## Runtime additions (Dart side)

- New `MockPaymentChannel` registered by `PlatformPaymentModule` when `Core.act.isMocked` is true. No-op Adapty replacement.
- New `Core.act.isMocked` flag on the Act mixin (symmetric with `isTest` / `isProd`).
- `PlatformCoreModule`, `PlatformCommonModule`, and `PlatformPermModule` register their real iOS channels under `isMocked` (previously only `isProd` triggered them — under `platformIsMocked` all platform channels fell through to in-memory / no-op stubs, so secure-storage writes dropped on the floor, HTTP calls threw, and `getPrivateDnsState` returned hardcoded ``unavailable``). With these, `FamilyActor.start` succeeds and the DNS-install wizard never opens.
- `PlatformPermActor._isPrivateDnsStateEnabled` (v6) and family's `PermActor._checkDns` (family) trust the iOS mock's ``enabled`` verdict without round-tripping the device-tag/alias URL the mock can't construct.
- New `common/lib/main_mocked_shared.dart` hosts the shared `seedDevAccount(Flavor)` helper + `MockedStart` class (was duplicated verbatim in both mocked entries). Seeds an active `JsonAccount` to secure storage on first launch so the paywall / first-account-create never fires; real `account.fetch()` overwrites on next refresh.
- `account.dart`: `_keyAccount` promoted to library-public `keyAccount` so the seed helper and `refresh.dart` use a single source of truth for the secure-storage key.
- `main-mocked-family.dart` sets `jsonUrl = https://family.api.blocka.net` (latent bug — family mocked entry was hitting the wrong base URL) and drops the now-stale `attachMockedDeps()` call.
- `common/lib/mocked-deps.dart` deleted (was the legacy static-mocks surface that the real-API path supersedes).
- `dev_seed.dart.template` + `.gitignore` entry: per-developer account IDs stay out of the repo; copy template once per checkout.

## iOS shell

`PrivateDnsServiceMock` now reports `active = true` from the start so `getPrivateDnsState` returns ``.enabled`` — matches what the real service would report on an already-onboarded device.

`FamilyMocked` target's `Sources` build phase drops four `NotFamily/Location/*.swift` entries that don't belong (matches what `FamilyDev` already excludes), fixing the pre-existing compile failure where `Container.plus` / `OpsGateway.niceName` didn't resolve.

## Appium harness

- `IOS_USE_SIM=1` mode reads the per-worktree sim UDID from `make -C ios sim-status`, routes the install step to `appium-install-{six,family}-mocked`, and skips device-only signing caps + `devicectl`-based WDA process inspection. Reuses the same `explore.mjs` JSONL surface for both device and sim.
- `app-targets.mjs#resolvePrimaryBundleId` returns `net.blocka.app` for both flavors in sim mode (the two Mocked Xcode targets share `PRODUCT_BUNDLE_IDENTIFIER`).
- `explorer-commands.mjs`: `ui.tap` accepts `{x, y}` for coordinate taps; new `ui.tapCenter` resolves an element by selector and taps its on-screen center via `mobile: tap`. Needed because Flutter widgets render as `XCUIElementTypeStaticText` without `Semantics(button: true)`, and `element.click()` targets the accessibility element but doesn't dispatch the underlying `GestureDetector` (real finger taps work because they're physical hit-tests).
- `sim.mjs`: new helper that resolves the worktree's sim from `sim-helpers.mk` output.

## Docs

`ios/SIMULATOR.md` — one-time `dev_seed.dart` setup, the "real API + system mocks" framing, the shared-bundle-id sim coexistence rule, and the `IOS_USE_SIM=1` Appium invocation.

## Test plan

- [x] `cd common && fvm flutter test --no-pub` — 207 tests, all pass.
- [x] Both kernels embed the correct mocked entry (`strings .../flutter_assets/kernel_blob.bin | grep main-mocked-*` returns `main-mocked-six.dart` / `main-mocked-family.dart` respectively, plus `main_mocked_shared.dart`).
- [x] V6 Appium tour: dismiss rate prompt + Safari extension sheet → reach the real V6 home (BLOKADA logo + power ring + Senaste 24h stats + Privacy Pulse / Avancerat tabs) → tap power → native "Annonsblockering är aktiv" action sheet → Privacy Pulse screen with donut chart + Toppdomäner / Senaste aktivitet sections.
- [x] FamilyMocked Appium tour: tap Aktivera → `activateCta` creates a real parent device on `family.api.blocka.net` → family home renders with the parent device cards + "Lägg till en enhet" CTA → tap a device → device-detail screen (STATISTIK + ENHETSINSTÄLLNINGAR + INTERNETKONTROLL sections — the surface #292's schedule editor will extend).
- [x] No DNS-install wizard, no Adapty / Mock-http errors in either sim log.

The pinned dev accounts referenced via `dev_seed.dart` (developer-local) need the right tier for each flavor — flip via the `account_activate` support tool as needed (V6 → cloud, Family → family).


balboah:
@codex please review

balboah:
@codex please re-review — committed `dev_seed.dart` with REPLACE_ placeholders so the codebase builds in a fresh checkout (your P1). Template file retired; SIMULATOR.md updated with the new edit-locally + untrack workflow.

balboah:
@codex small follow-up push (53639f40) tightens the SIMULATOR.md note about local-edit handling — jj doesn't have a clean equivalent of `git update-index --skip-worktree`, so the doc now reflects that honestly rather than over-promising a one-liner.

balboah:
@codex small doc-only follow-up at 0c99ccf0: added a worktree-friendly alternative to the submodule init step in AGENTS.md (cp -cR from main checkout for APFS speed). No code change.

balboah:
Closing — will reopen with a revised approach (dropping dev_seed.dart in favor of dart-define + optional .env). Branch and worktree retained locally.

PR #1056, opened 2026-05-20
smoke: detect stale Adapty fallback paywalls bundle

kar:
## Summary

- Adds an Appium smoke spec `automation/appium/wdio/src/specs/smoke/adapty-fallback.spec.ts` that pulls the in-app share-log after a cold start and asserts the Adapty SDK did not reject the bundled fallback paywalls JSON.
- Detection lives in a small helper `automation/appium/wdio/src/flows/adapty.ts` matching four signatures of error code 2006 (`Failed setting fallback`, `Decoding Fallback Paywalls`, `fallback paywalls version is not correct`, `code: 2006`).
- Reuses the existing `pullRecentDeviceLog` library from `automation/device/lib/log.mjs` — no `idevicesyslog`/OSLog plumbing.

## Why now

Error 2006 currently rots silently: the Dart catch in `common/lib/src/features/payment/domain/adapty.dart:39` swallows the `PlatformException` with a debug-level `ignore` log, so the bundled JSON can go stale relative to the SDK version for months without anyone noticing. This smoke moves that signal into CI.

## Expected behavior on merge

**This smoke is expected to turn red on `main` after merge** until the companion PR (`adapty-fallback-refresh`, refreshes the three bundled JSON files from the current Adapty Dashboard export) lands. Merging the detection first lets us confirm CI actually catches the issue before applying the fix.

## Test plan

- [ ] CI runs the new spec on the self-hosted iOS runner.
- [ ] CI fails with a message pointing at the three JSON paths that need refreshing.
- [ ] Companion PR (`adapty-fallback-refresh`) merges; smoke turns green.

### DIFF
--- a/automation/appium/wdio/src/flows/adapty.ts
+++ b/automation/appium/wdio/src/flows/adapty.ts
@@ -1,0 +1,49 @@
+import { driver } from "@wdio/globals";
+import { expect } from "chai";
+
+const STALENESS_PATTERNS: RegExp[] = [
+  /Failed setting fallback/i,
+  /Decoding Fallback Paywalls/i,
+  /fallback paywalls version is not correct/i,
+  /code:\s*2006\b/i,
+];
+
+type PullRecentDeviceLog = (options: Record<string, unknown>) => Promise<{ text: string }>;
+
+export async function assertAdaptyFallbackHealthy(bundleId: string): Promise<void> {
+  // Give the catch in common/lib/src/features/payment/domain/adapty.dart:39
+  // time to flush the failure line to the in-app share-log.
+  await driver.pause(1500);
+
+  const caps = driver.capabilities as Record<string, unknown>;
+  const udid = (caps["appium:udid"] ?? caps["udid"]) as string | undefined;
+  if (!udid) {
+    throw new Error("Adapty fallback check: no UDID in wdio capabilities");
+  }
+
+  const logModule = (await import("../../../../device/lib/log.mjs")) as {
+    pullRecentDeviceLog: PullRecentDeviceLog;
+  };
+
+  const { text } = await logModule.pullRecentDeviceLog({
+    bundleId,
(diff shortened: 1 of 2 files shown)


kar:
Closing in favor of a simpler static repo-root check.

## What I tried

1. **`pullRecentDeviceLog` (devicectl, app-group container)** — fails on CI and locally with CoreDevice `Code 1002 "No provider was found"`; needs a developer provisioning provider that authorizes enumerating `group.net.blocka.app`. Setup we don't want to take on.
2. **`driver.getLogs("syslog")`** — worked on my iPhone Maxi (captured the exact `code: 2006` line within seconds of cold start), but the CI run's same call on `iPhone-Tomasz` returned 100 entries all from a single second of `testmanagerd`/`SpringBoard` noise, no Adapty/Flutter lines at all. `appium-ios-device`'s lockdown syslog stream is unreliable across hosts in a way I couldn't diagnose remotely.
3. **`mobile: pullFile` of a runtime marker file** — Appium's real-device pullFile only supports the `documents` container, which needs `UIFileSharingEnabled = YES` in Info.plist. That's a real product change (Blokada would show in iOS Files.app), out of scope.

## What works instead

The bundled fallback JSONs encode their format version at `meta.version`:

- Broken (currently on main): `"meta":{"version":8,...}` — SDK 3.15.7 rejects with code 2006.
- Fixed (#1057): `"meta":{"version":9,...}` — SDK 3.15.7 accepts.

A 20-line Node script in CI that parses each fallback JSON and asserts `meta.version >= 9` catches the exact same staleness with no device, no Appium, no entitlements. When Adapty bumps the format alongside an SDK upgrade, the expected version gets bumped in the same PR that refreshes the JSONs.

Follow-up PR coming with the script + GitHub Actions workflow. #1057 (the JSON refresh itself) is unchanged.

PR #1061, opened 2026-05-25
fix(payment/ios): pin adapty_flutter to 3.11.4 to restore paywall

balboah:
## Summary

The iOS paywall (six/cloud flavor) white-screens, dropping new subscriptions to near-zero since the 26.2.x rollout.

Root cause: **adapty_flutter 3.15.0** made `ProductReference.fromJsonValue` read `access_level_id` and `product_type` as non-null `String`. Adapty's `get_paywall` response for placement `primary` omits both, so the Dart decode throws error `10002` ("type 'Null' is not a subtype of type 'String'") and the paywall never renders. Confirmed via on-device verbose logs + the raw `get_paywall` payload, and a published-version bisect: versions jump `3.11.4 -> 3.15.0`, and `3.11.4` is the newest release whose decoder ignores those fields.

This pins `adapty_flutter` to **3.11.4**.

## Changes
- `common/pubspec.yaml` (+ lock): `adapty_flutter` `3.15.5 -> 3.11.4`.
- `common/lib/src/features/payment/domain/adapty.dart`: revert iOS rendering to `AdaptyUI().createPaywallView().present()` (the embedded `AdaptyUIPaywallPlatformView` sheet is 3.15-only).
- Remove `paywall_sheet.dart` and its `part` line; drop now-unused imports.

## Scope / notes
- **iOS-only.** Android uses the separate native `io.adapty 3.15.2` SDK and is untouched. The bundled fallback assets are left as-is.
- **UX:** the iOS paywall reverts from the embedded bottom-sheet to the full-screen native modal (as it was pre-26.2.2).
- **Interim pin.** Revert to 3.15.x once Adapty makes `access_level_id`/`product_type` optional in the get_paywall decoder; that restores the embedded sheet (recover `AdaptyEmbeddedPaywallSheet` from git history).

## Verification
- Device build (run-six, iPhone 17 Pro, iOS 26.5): paywall renders correctly.
- Clean control on 3.15.5 (fresh pods + fresh install) reproduced the white screen, ruling out pod-mismatch and caching.
- `dart analyze` of the payment domain: clean.

## Test plan
- [ ] Paywall renders on a six/iOS build (confirmed locally)
- [ ] A real purchase completes end-to-end
- [ ] Restore purchases works
- [ ] No regression on Android (unchanged, native SDK)

Ref: blokadaorg/issue-tracker#311


kar:
I think we can ignore the other CI smoke test (red), it's just an experiment.

balboah:
@codex please re-check.

Addressed the P1 (single-use `AdaptyUIPaywallView`): `showPaymentScreen` now clears the cached `_paymentView` (and its placement id) and rethrows if `present()` throws, so a spent view is never re-presented and the next open recreates it.

Note: the failing `smoke` check is the DNS-onboarding power-toggle test (`waitForProtectionActive`), unrelated to this payment-only change — re-running with this push.

balboah:
@codex hello? check update

balboah:
Codex seems rather broken after the 1st review (no verdict on the re-check). Local review looked good; addressed its one finding (await the paywall-view `dismiss()`). Ready for your review @kar.

kar:
yep, lgtm. merging and making the build

PR #1066, opened 2026-05-29
feat(family): scheduled filtering on kid devices

balboah:
## Summary

Delivers family scheduled filtering on kid devices (closes blokadaorg/issue-tracker#292 "family: scheduled filtering on kid devices").

A parent can now define weekly time windows on a kid's device that switch the device's active profile during those windows. Outside any window, the device falls back to its base profile. A device-wide pause toggle dims dormant rules so the dormant state is visible.

### Model + storage
- `TimeWindowModel` with explicit wraparound semantics for windows that cross midnight.
- `RuleModel` + `ScheduleModel` with client-side validation.
- `JsonDevice` round-trips the schedule + IANA timezone (resolved via a platform channel) through `put` payloads.
- `ScheduleActor` with pure transforms and save.
- School + Bedtime templates seeded on first device setup.

### Rule editor (full-page, iOS-style)
- Profile chip row with horizontal-scroll, dual-edge soft fade, long-press quick-delete via an anchored popover, +New mini-flow (rename → blocklists → select).
- Day toggles with weekdays / weekends / every preset chips.
- Compact times list with a soft cap, wrap-to-next-day hint, and a counter folded into the section label.
- Override-hint paragraph clarifying the rule lives on top of the device's base profile, which is named in the sentence rather than shown as a redundant chip.

### Profile editor + shared nav
- `ProfileAvatar` (deterministic palette colour, grapheme-cluster initial letter so emoji aliases render as the emoji not a missing-glyph "?") replaces ad-hoc icon call sites.
- Shared `BackTitleNavBar` so the rule editor and profile editor both use iOS-standard navigation (back chevron labelled with the previous screen's title, centred current-screen title, trailing action).
- Glass effect on the top nav and the profile editor's bottom action dock (translucent fill behind a `BackdropFilter`, content scrolls under both surfaces).
- Bottom edges of cards swipe-to-delete; rule rows dim when the device-wide schedule toggle is off.

### Schedule section on Device detail
- Positive-on toggle ("Use schedule") with subtitle that adapts to the active/paused state.
- Rule list with profile colour pill, day summary, times summary.
- Pause dims the rule rows so the user can see at a glance that nothing will fire.

### i18n + tooling
- English keys in the shared SoT, mirrored to iOS `Ui.strings` and Android `strings_ui.xml`; Swedish translations added in the same passes.
- Appium harness gains `ui.longPress`, `ui.swipe`, `ui.alert`, `ui.tapCenter` fixes, and an `IOS_UDID` override knob to script the new gestures end-to-end.

## Verification
- `fvm flutter analyze` in `common/`: clean (six pre-existing warnings outside this branch).
- `fvm flutter test --no-pub` in `common/`: 291 / 291 passing.
- FamilyMocked iOS sim build green; manual walkthrough through device detail → rule editor → profile editor (incl. emoji-named profile) on iPhone 17 sim.


balboah:
@codex please review this PR. History was rewritten from 56 small iterative commits into 7 themed commits via forward-construction; cumulative diff matches the original branch tip apart from intentionally dropped `.superpowers/brainstorm/` artifacts that were accidentally committed before `.gitignore` caught up. Common test suite (291/291) and `flutter analyze` are clean; FamilyMocked iOS sim build verified locally.

balboah:
@codex addressed: moved `_seedScheduleForNewDevice` call from `DeviceActor.reload` (controller/parent device path) into `DeviceActor.addDevice` (linked kid-device path), so seed lands on the kid device whose base profile the parent is restricting. Doc comment updated to reflect kid-device semantics. Folded into the model+storage commit; PR still 7 commits.

balboah:
@codex addressed both round-2 findings:

1. **P2 seed schedules for newly linked kid devices**: same as round-1 P1, already fixed — `_seedScheduleForNewDevice` is called from `DeviceActor.addDevice` (linked-kid path), not `reload` (controller/parent path).
2. **P2 remove committed local brainstorm artifacts**: confirmed and removed. Those 9 `.superpowers/brainstorm/` files (HTML + server.pid + server-stopped) had slipped into the first commit via a working-tree snapshot during the rewrite. Folded the deletion back into the introducing commit so they're gone from the chain. The `.gitignore` rule remains so new artifacts stay out.

Tests 291/291, analyzer clean. Same 7-commit shape; head now `1845e4088531`.

balboah:
@codex

balboah:
@codex addressed round-3 P2 (best-effort schedule seeding on `addDevice`):

Wrapped `_seedScheduleForNewDevice` in a try/catch inside `DeviceActor.addDevice`, log the error, and continue. The device POST has already succeeded server-side at that point, so a seed failure no longer:

- tanks the link flow (QR token is still returned to the parent's link sheet), or
- orphans the just-created device if the parent retries.

Parent can configure the schedule from the Schedule section once the link completes. Folded into the model+storage commit; still 7 commits. Tests 291/291, analyzer clean. Head now `6b3498c1943e`.

balboah:
@codex

balboah:
@codex addressed round-4 P2 (Localize schedule day summaries):

`daysSummary` now routes the three canonical preset branches through the existing `family schedule days summary {weekdays,weekends,every}` i18n keys, so Swedish renders `Mån–Fre`, `Lör–Sön`, `Varje dag` and English keeps `Mon–Fri`, `Sat–Sun`, `Every day`. Pure-transform tests now assert the keys (i18n returns the key verbatim in test env) and the schedule-section widget test was updated to match.

Note: custom day-set summaries (e.g. `[Mon, Wed, Fri]`) still render in English because the individual short day labels don't yet have i18n keys — that needs a fresh key surface and is intentionally out of scope for this PR; can do as a follow-up.

Folded into the model+storage and Schedule-section commits respectively; still 7 commits. Tests 291/291, analyzer clean. Head now `1adec948faa3`.

balboah:
@codex

balboah:
@codex addressed round-5 P3 (Localize the multi-window summary suffix):

`windowsSummary` now routes the "+N more" suffix through the `family schedule rule row more` i18n key, so non-English locales render their own translation (Swedish: `, +N till`) instead of mixed-language text. The key already carries the leading `, +` plus the `%s` count placeholder, so the call is `'family schedule rule row more'.i18n.withParams(remaining.toString())`. The pure-transform test was updated to assert the structure (first two windows verbatim, suffix contains the localised key surface) since the test env returns the key verbatim.

Folded into the model+storage commit; still 7 commits. Tests 291/291, analyzer clean.

balboah:
@codex

balboah:
@codex addressed both round-6 P2 findings:

1. **Resolve local timezone on first schedule saves** — `ScheduleActor.saveSchedule` now keys the first-save timezone decision off `device.schedule == null` rather than `device.timezone` being null. If the backend has already defaulted the legacy device's `timezone` field to UTC, the first interactive save now resolves the platform's IANA zone instead of preserving the default. Subsequent saves still preserve the existing `device.timezone` unless the caller overrides it. Doc comment updated.
2. **Treat parent-profile legacy rules as ineligible** — `_RuleEditorSheetState.initState` now clears `_profileId` whenever the seed value isn't in `_eligibleProfiles`, not just when it equals the device base. Catches any rule whose existing `profileId` points at a parent-template profile, a deleted profile, or any other ineligible value; the Save gate then forces the user to pick an eligible chip instead of silently re-persisting the broken state.

Folded into the model+storage and Schedule-section commits respectively; still 7 commits. Tests 291/291, analyzer clean.

balboah:
@codex

balboah:
@codex addressed round-7 P2 (Commit seeded device response before publishing it):

`_seedScheduleForNewDevice` now returns the `JsonDevice` carrying the persisted seed (it always had it as the result of `_devices.changeSchedule`, just wasn't returning it). `addDevice` captures that and publishes the seeded record into both the local cache and the returned `LinkingDevice`. If the seed fails, the local copy keeps the pre-seed record (so the link flow still hands back a QR and the parent can configure the schedule from the Schedule section) — matching the best-effort behaviour from round 3. No more empty-schedule fallback overwriting the backend's seeded rules on a subsequent edit.

Folded into the model+storage commit; still 7 commits. Tests 291/291, analyzer clean.

balboah:
@codex

balboah:
@codex addressed round-8 P2 (Clean up seeded profiles when linking is cancelled):

Deferred the seed entirely. `DeviceActor.addDevice` no longer creates the School/Bedtime profiles or writes the schedule; it just POSTs the device and returns the `LinkingDevice`. A new `DeviceActor.seedScheduleForLinkedDevice` runs the seed (same best-effort error handling) and is now invoked from `LinkActor._finishLinkDevice` after the heartbeat-confirmed link completes.

Result: when the parent cancels via `LinkActor.cancelLinkDevice`, no seed has run yet, so there are no School/Bedtime profiles to clean up and `_device.deleteDevice` is the only rollback needed (as before). When the link succeeds, the seed runs once and the local cache picks up the seeded `JsonDevice` (preserving the round-7 fix).

Folded into the model+storage commit; still 7 commits. Tests 291/291, analyzer clean.

balboah:
@codex

balboah:
@codex two findings from round 9:

**P1 — Avoid reseeding relinked devices** — fixed. `seedScheduleForLinkedDevice` now early-returns when `device.schedule != null`, so the relink path (`LinkingDevice(device: existing, relink: true)`) cannot overwrite the parent's customised rules with the locked seed. Guarding on `device.schedule` rather than the transient `relink` flag also covers any future path that hands a previously-configured device through this entry point.

**P2 — Don't persist the parent's timezone for child schedules** — acknowledged but punted. The architectural fix needs the kid device's IANA zone to flow back from the kid device itself; the current Family flow has no infrastructure for that (the kid runs the DNS profile, not the Flutter app, so `platformTimezone()` only reports the parent's phone's zone). The current code is still the best guess available for the cross-TZ household case, and the round-6 guard (resolve from platform on first save) is still needed to overcome the backend's default-UTC pre-fill. Tracking this as a follow-up; same-household families (the dominant case) are unaffected.

Folded P1 fix into the model+storage commit; still 7 commits. Tests 291/291, analyzer clean.

balboah:
@codex

kar:
Great stuff. Some things I could catch:

- My profile was not applied to the child device despite the rule was valid for current local time. In home screen it still shows the default profile as the active one. Re-toggling did not apply it. But not sure if it's already implemented?
- "Use schedule switch" should probably have some pill-like background , looks a little bit out of place to the eye for me. minor

IMG_4917





- Rule editor itself is a little inconsistent with the rest of UI when it comes to visual hierarchy - font sizes and placement of subtitles / brief texts - minor but could be easily improved. could try to ask claude to get another look on other screens to get inspired.
- Guarda(r) button seems to wrap (nav bar)
- The days box width is shrinking depending on button selections, should be full width always. also buttons jump around depending if they are selected there or not

IMG_4918






- Profile creation dialog - doesn't trim empty space of profile name, inconsistent styling - we don't use icon for buttons in navbar (edit - pen). Done and Delete profile constantly displayed at the bottom will steal vertical space on some layouts (probably horizontal ipad, and small phones). would be better to put the trash icon on same line as the CTA. 
- Profile icon badge in the new UI is inconsistent with the ones is the quick profile selector (home screen -> swipe on card -> profile)

Captura 2026-06-02 a las 9 56 39

balboah:
> In home screen it still shows the default profile as the active one

to clarify this feedback, you expect the app to compute currently active profile in home view client side according to configured rules, instead of showing the default profile?

> My profile was not applied to the child device despite the rule was valid for current local time

When you say it was not applied, do you mean activity log did not show blocks? This is because backend isn't deployed to support this feature.

kar:
Yeah I just did not notice any place in UI that tells me what profile is currently active for that device, according to schedule. I assumed it would show in the home screen. If that shows only the default profile, this may be confusing for users.

PR #1085, opened 2026-06-05
Inline Dependabot auto-merge workflow for public repo

kar:
## Problem

The Dependabot auto-merge wrapper called a reusable workflow in the **private** repo `blokadaorg/workflows`:

```yaml
uses: blokadaorg/workflows/.github/workflows/dependabot-auto-merge.yml@main
```

GitHub does not allow a **public** repository to call a reusable workflow that lives in a **private** repository — even with org-level sharing enabled. So every run failed at the validation stage before any job started (0 jobs, 0s, "This run likely failed because of a workflow file issue"). The same reusable workflow runs fine in the org's private repos (`api`, `blockadns`, …); `blokada` is the only public consumer.

## Fix

Vendor (inline) the reusable workflow's four jobs directly into the caller, matching the self-contained pattern already used by `blokadaorg/landing-github-pages` (which intentionally avoids depending on private workflow code for the same reason).

Deltas from the upstream reusable workflow:
- `on: workflow_call` → `on: workflow_run` (CI / completed)
- `inputs.allowed_files_regex` → literal regex (char-for-char identical to the previous caller value)
- `review` gate keeps the caller's fork-safety guard (`head_repository.full_name == github.repository`) plus the `DEPENDABOT_AUTOMERGE_PAUSED` kill switch
- secrets resolve directly from org/repo scope instead of via `secrets: inherit`

All safety-critical logic is reproduced verbatim: major-bump detection, the changed-file allowlist, the `github_actions`-ecosystem human-review carve-out, the stale-head guard, and the prompt-injection-hardened Claude review.

## Notes

- No secret setup needed — the same secrets already reached the workflow via `secrets: inherit`, so they're already configured for this repo.
- `workflow_run` workflows only run from the **default branch**, so this fix takes effect once merged to `main`; the next Dependabot PR will exercise it.

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -164,6 +164,11 @@ jobs:
 
       - name: Run Claude review
         if: steps.resolve.outputs.pr_number != '' && steps.precheck.outputs.verdict == 'PASS'
+        # A hard error from the action (transient API failure, etc.) must not
+        # fail the whole review job — otherwise Parse decision is skipped and
+        # the ABSTAIN fallback never fires. Let it continue; a missing
+        # /tmp/decision.txt is then parsed as ABSTAIN and surfaced to humans.
+        continue-on-error: true
         uses: anthropics/claude-code-action@v1
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
@@ -172,7 +177,7 @@ jobs:
           allowed_bots: 'dependabot[bot]'
           additional_permissions: |
             actions: read
-          claude_args: '--allowed-tools "Bash(gh pr view:*),Bash(gh pr diff:*),Bash(cat:*),Write"'
+          claude_args: '--allowed-tools "Bash(gh pr view:*),Bash(gh pr diff:*),Write"'
           prompt: |
             You are reviewing Dependabot PR #${{ steps.resolve.outputs.pr_number }}
             in ${{ github.repository }}. CI has passed. Pre-checks have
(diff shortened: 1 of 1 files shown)


kar:
Reviewed the Copilot feedback — all three points are valid and addressed in c6b37cb8.

**1 + 2 (Claude-step failure breaks the ABSTAIN fallback) — valid, fixed together.**
Confirmed the Actions semantics: a step `if:` with no status function carries an implicit `success()`, so a hard error from `Run Claude review` would skip `Parse decision` and fail the `review` job. With `decision` then empty and `needs: review` requiring success, neither `merge` nor `annotate-abstain` would run — so no ABSTAIN comment despite that job existing precisely to surface automation failures. Added `continue-on-error: true` to the Claude step and `if: always() && steps.resolve.outputs.pr_number != ''` to `Parse decision`. A missing `/tmp/decision.txt` now degrades to ABSTAIN and actually fires the annotation, which is the intended design.

**3 (`Bash(cat:*)` in allowed-tools) — valid, removed.**
The prompt restricts evidence gathering to `gh pr view`/`gh pr diff` and output to `/tmp/decision.txt` via `Write`, so `cat` was unused. In a prompt-injection-sensitive step (untrusted upstream release notes) it only widened blast radius for arbitrary file reads (e.g. `/proc/self/environ`). Dropped it.

Heads-up on divergence: this file is a vendored copy of the reusable workflow in the private `blokadaorg/workflows` repo. These three fixes now make the inlined copy diverge from upstream — they should be mirrored back into the upstream reusable workflow (and the other private-repo consumers benefit too), otherwise a future re-vendor would silently revert them.

PR #1098, opened 2026-06-10
feat(family): scheduled filtering on kid devices

balboah:
Implements the app side of family scheduled filtering on kid devices (blokadaorg/issue-tracker#292). Supersedes #1066, which was closed; the work is consolidated here into a clean, commit-by-commit reviewable stack.

## What this delivers

A parent can give a kid device a default profile, carve stricter time-based windows with schedule rules (switch profile, or block internet), and apply a temporary manual override (pause filtering, or block internet) that auto-reverts. The device-detail screen makes the effective state legible at a glance.

## Highlights

- Model, storage, and a display-only resolver for the precedence stack (manual override, then schedule rule, then default profile), with timezone-aware window math documented as a client-side approximation.
- Schedule section and rule editor on Device detail, profile editor refresh with full-page navigation, shared ProfileAvatar and nav primitives.
- "Right now" readout: one tappable row stating what is in effect, which setting decided it, and until when. Tapping opens a change-now action sheet (Resume while overriding, then Pause filtering, or Block internet) that chains into the duration sheet. A single in-control bar travels between the readout, the active schedule rule, and the default profile, so exactly one is shown at a time.
- English and Swedish strings; design specs under docs/superpowers/specs.

## Testing

- Full common test suite green; widget tests cover the readout states, the sheet flow, override suppression of the schedule marker, and the schedule and resolver units.
- Verified end to end on the FamilyMocked simulator: section order, the change-now sheet, applying Block internet for 1 hour (readout flips to the override state with its caption and bar), and Resume restoring the schedule and default attribution.

## Notes for review

- Reviewable commit by commit; each commit is one logical phase.
- App side only; backend wiring is sequenced separately under the same issue.


balboah:
@codex please re-check. Addressed the P2 on the home card: a firing schedule block rule now shows the "No internet" label instead of falling back to the default profile name (new `activeCardLabelKey` helper + tests), squashed into the device-detail-redesign commit.

balboah:
@codex please re-check. Addressed the P3: `changeSchedule`'s optimistic draft now carries `modeUntil` so a schedule edit during a bounded override no longer flips it to "indefinite" before the API reconciles (folded into the mode_until model/data commit). Also cleaned the PR: dropped an incidental Firebase Podfile.lock bump and removed committed Claude/superpowers planning docs (now gitignored).

balboah:
@codex please re-check. Addressed the latest review comments on 590528fc: the home card label now uses the shared effective-state resolver so manual Block/Pause overrides outrank schedule/default labels, and rename/profile-change optimistic drafts now preserve modeUntil. Added regressions for active/expired home-card overrides and pending-API optimistic drafts. Local verification: make test passed.

balboah:
@codex please re-check the latest P2 about explicit null mode_until in mode updates. I verified this against the v3 API behavior: when a PUT includes an explicit mode, the handler always applies the mode_until option with the request pointer, and an omitted mode_until is the nil clear signal. That is intentional for both Resume (mode=on, no mode_until) and switching a timed block/pause to an indefinite override (mode=blocked/off, no mode_until). The focused API regression TestDeviceHandlerV3ModeUntilClearedOnModeChange covers both cases and passes via make test RUN=TestDeviceHandlerV3ModeUntilClearedOnModeChange PKG=. So the app payload shape here is intentional: include mode, omit mode_until to clear any previous expiry.

PR #1118, opened 2026-06-16
build(android): migrate to AGP 9.2.1 + built-in Kotlin + compileSdk 37

kar:
## Why

The android-gradle Dependabot group (**#1100**) can't build on AGP 8.13.0: `androidx.core:core(-ktx):1.19.0` requires **AGP ≥ 9.1.0 + compileSdk 37**, and Adapty's Compose-based paywall UI needs a Compose dependency that AGP 9 makes the consumer supply. This PR is the **toolchain prerequisite** so that group (and future androidx bumps) can land. It **supersedes the standalone AGP bump #1080**.

Scope is deliberately toolchain-only — the app-dependency bumps from #1100 (androidx core/activity/lifecycle/etc., `adapty-bom 3.17.1`) are **not** here; recreate #1100 on top once this merges.

## What AGP 9 forced (each verified by building both flavors)

- **AGP 8.13.0 → 9.2.1**, **compileSdk 35 → 37** (needs SDK platform `android-37.0`).
- **Built-in Kotlin is mandatory at 9.2.1** — the external `kotlin-android` plugin is rejected (the `builtInKotlin=false` / `newDsl=false` escape hatches no longer work at 9.2.1), so it's removed and Kotlin comes from AGP. The Kotlin toolchain is aligned to a built-in-Kotlin-compatible set: kotlin-gradle-plugin & kotlin-stdlib **2.4.0**, KSP **2.3.9**, serialization **2.4.0**; safe-args gradle plugin **2.8.9 → 2.9.8** (the 2.8.9 plugin fails to detect AGP 9).
- **Source sets:** built-in Kotlin compiles `kotlin.srcDirs`, not `java.srcDirs`. The custom Kotlin dirs (`src/engine/kotlin`, `wireguard-android/ui`, `src/mock/kotlin`) are now registered as `kotlin.srcDirs` (the default `src/<flavor>/kotlin` dirs are auto-included).
- **Proguard:** AGP 9 drops `getDefaultProguardFile('proguard-android.txt')` → `proguard-android-optimize.txt`.
- **Compose for Adapty:** `io.adapty:android-ui` pulls `androidx.compose.*` without versions; under AGP 9 the consumer must pin them, so a **Compose BOM** is added to the `six` flavor. We don't author Compose — the BOM only versions Adapty's prebuilt UI. (This is AGP-9-driven, not Adapty-version-driven — it reproduces on the current `adapty-bom 3.15.2`.)

## Validation

Built locally on this branch (current main deps), matching the CI `Android assemble` step (`make -C android apk-family-debug apk-six-debug`):

- ✅ `apk-six-debug` — **BUILD SUCCESSFUL** (six = Adapty + WireGuard + Firebase)
- ✅ `apk-family-debug` — **BUILD SUCCESSFUL**
- Flutter common AAR (`regen-android`) unaffected; iOS/Flutter-analyze CI jobs untouched.

## Operator notes

1. **Self-hosted runners need SDK platform `android-37.0`** installed (`sdkmanager "platforms;android-37.0"`) — it's API 37 / minor 0 (AGP 9 minor-SDK scheme). Builders without it will fail at `checkAarMetadata`.
2. **Release builds now use `proguard-android-optimize.txt`** (R8 optimizations on; the old default carried `-dontoptimize`). CI only assembles debug — smoke-test a **release** build before shipping.

## Follow-ups
- Close **#1080** (AGP 9.2.1 standalone) — absorbed here.
- After merge, `@dependabot recreate` **#1100** so it carries only the app-lib bumps + `adapty-bom 3.17.1`; then it needs on-device Adapty purchase/restore validation.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -12,7 +12,6 @@
 
 plugins {
     id 'com.android.application'
-    id 'kotlin-android'
     id 'com.google.devtools.ksp'
     id 'androidx.navigation.safeargs.kotlin'
     id 'org.jetbrains.kotlin.plugin.serialization'
@@ -25,7 +24,7 @@ kotlin {
 
 android {
     namespace "org.blokada"
-    compileSdk 35
+    compileSdk 37
 
     kotlin {
         jvmToolchain 17
@@ -81,7 +80,7 @@ android {
             }
 
             minifyEnabled true
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
 
             ndk {
                 abiFilters 'armeabi-v7a', 'arm64-v8a'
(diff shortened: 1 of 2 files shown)


kar:
Addressed all three Copilot comments in 83f2c9ae — all were valid; verified by rebuilding `apk-six-debug` + `apk-family-debug` (both green):

1. **Removed the `kotlin-gradle-plugin` buildscript classpath.** Confirmed redundant under built-in Kotlin: `buildEnvironment` shows the serialization (2.4.0) and KSP (2.3.9) plugins already pull `kotlin-gradle-plugin` to **2.4.0** (AGP's bundled 2.2.10 → 2.4.0 via conflict resolution), so the toolchain stays at Kotlin 2.4.0 with no skew.
2. **Removed `six.java.srcDirs += 'src/engine/kotlin'`** — compiles via `kotlin.srcDirs` (dir is 0 `.java`).
3. **Removed `six.java.srcDirs += 'wireguard-android/ui/src/main/java'`** — same (0 `.java`).

Also reverted a stray `debug.kotlin.srcDirs += 'src/mock/kotlin'` (that dir is empty).

PR #1135, opened 2026-06-19
fix(ios): don't re-sign Flutter/App xcframeworks (fixes broken release archive)

kar:
## Problem

The **Release from tag** archive for `26.2.18/family/ios` failed ([run 27819630775](https://github.com/blokadaorg/blokada/actions/runs/27819630775/job/82330476324)):

```
❌ error: signature-collection failed: The operation couldn't be completed.
   (SWBUtil.CodeSignatureInfo.Error error 0.)
   SignatureCollection .../Release-iphoneos/Flutter.xcframework-ios.signature
   ProcessXCFramework .../common/build/ios-framework/Release/Flutter.xcframework ...
** ARCHIVE FAILED **
```

(The `export_method mismatch` lines below it in the log are gym's post-failure fallback noise, not the cause.)

**Root cause:** the ITMS-91065 signing fix (#1133) used `find ... -name '*.framework'` and re-signed **every** framework under the `ios-framework` output — including Flutter's own first-party `Flutter.framework` and `App.framework`, which already ship signed (`io.flutter.flutter` / `io.flutter.flutter.app`). Re-signing those breaks Xcode's `ProcessXCFramework` → `SignatureCollection` step on `Flutter.xcframework`.

**Why it merged green:** the PR `CI` workflow is build + lint only and never runs the gym Release archive / `sign-ios-frameworks`. The script's first real exercise is the `Release from tag` workflow — so the regression only surfaced at release time. The previous release (`26.2.17`, pre-#1133) archived fine.

The signing log confirms the asymmetry — only the plugins came out unsigned on the device slice; Flutter/App were `replacing existing signature`:

| device slice | state out of `flutter build ios-framework` |
|---|---|
| `sqflite_darwin`, `path_provider_foundation`, `shared_preferences_foundation`, `Adapty*` | **unsigned** → genuinely need signing (ITMS-91065) |
| `Flutter`, `App` | **already signed** → must NOT be re-signed |

## Fix

- **`scripts/sign-ios-frameworks.sh`** — exclude `Flutter`, `App`, and the static link-only `FlutterPluginRegistrant` xcframeworks from the signing loop; keep signing the unsigned third-party plugins ITMS-91065 actually flags. (`FlutterPluginRegistrant` is link-only — verified it appears only in the Link phase of `project.pbxproj`, never an Embed phase — so it never lands in the bundle and ITMS-91065 can't fire on it.)
- **`scripts/verify-ios-ipa-signatures.sh`** (new) — post-archive gate: cracks the produced `.ipa` and asserts every embedded `*.framework` (incl. Flutter/App) is signed by the Apple Distribution identity, and that Flutter/App are actually present. Fails the build **before** upload rather than at App Store Connect. Override the expected authority with `IOS_EXPECTED_SIGN_AUTHORITY` (set empty to check presence only).
- **`Makefile`** — run the verifier after the host archive in `build-ios` / `build-ios-family` / `build-ios-six`.

This narrows the script to match what its own header comment already claims ("the add-to-app Flutter **plugin** xcframeworks") — the plugin signing that fixes the rejection is fully retained.

## Testing

- `sign-ios-frameworks.sh` prune logic tested against a synthetic tree: signs exactly the 9 plugin xcframeworks (both slices), excludes Flutter/App/FlutterPluginRegistrant.
- `verify-ios-ipa-signatures.sh` tested end-to-end with synthetic IPAs: passes when all embedded frameworks are validly signed; fails on an unsigned framework, a missing required framework, a wrong/ad-hoc signing identity, and aggregates failures across multiple IPAs.
- Both scripts parse and run under stock macOS `/bin/bash` 3.2 (CI runner); no bash-4 features.
- `make -n` confirms the verifier is wired with the correct IPA paths (`ios/Blokada Family.ipa`, `ios/Blokada 6.ipa`).

## Known follow-up (not in scope)

`release-tag.yml` tars `ios/*.ipa` (glob) while building a single flavor. It's safe today because `actions/checkout` runs `git clean -ffdx` and the IPAs are gitignored, but a stale opposite-flavor IPA could ship if `clean: false` were ever set. Worth an explicit `rm -f ios/*.ipa` or tarring the specific flavor in a separate change.

## Operator action required

Merge this **before** re-cutting the `26.2.18` iOS release tag — the current release build is broken without it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,18 +80,21 @@ build-ios:
 	$(MAKE) -C common/ build-ios
 	$(MAKE) sign-ios-frameworks
 	$(MAKE) -C ios/ build
+	./scripts/verify-ios-ipa-signatures.sh "ios/Blokada Family.ipa" "ios/Blokada 6.ipa"
 
 # Build ios family .ipa from scratch (release)
 build-ios-family:
 	$(MAKE) -C common/ build-ios
 	$(MAKE) sign-ios-frameworks
 	$(MAKE) -C ios/ build-family
+	./scripts/verify-ios-ipa-signatures.sh "ios/Blokada Family.ipa"
 
 # Build ios six .ipa from scratch (release)
 build-ios-six:
 	$(MAKE) -C common/ build-ios
 	$(MAKE) sign-ios-frameworks
 	$(MAKE) -C ios/ build-six
+	./scripts/verify-ios-ipa-signatures.sh "ios/Blokada 6.ipa"
 
 # Code-sign the unsigned Flutter plugin xcframeworks before the host archive
 # embeds them, so commonly-used SDKs (sqflite, path_provider, ...) ship with a
--- a/scripts/sign-ios-frameworks.sh
+++ b/scripts/sign-ios-frameworks.sh
@@ -51,11 +51,27 @@ if ! security find-identity -v -p codesigning 2>/dev/null | grep -qF "$IDENTITY"
 	exit 1
 fi
 
+# Do NOT sign Flutter's own first-party outputs. Flutter.framework and
+# App.framework come out of `flutter build ios-framework` ALREADY signed
+# (io.flutter.flutter / io.flutter.flutter.app) and are not on the unsigned-
+# plugin list ITMS-91065 flags. Re-signing them here breaks the host archive:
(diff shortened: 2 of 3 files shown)


kar:
@codex review this

PR #1137, opened 2026-06-19
fix(ios): sign plugin xcframework bundles with --timestamp (real ITMS-91065 fix)

kar:
## TL;DR

ITMS-91065 kept rejecting `sqflite_darwin` across **26.2.18 → 26.2.20** even though the embedded framework was validly code-signed. By inspecting the actual rejected IPA I found ITMS-91065 reads a *different* artifact than we'd been fixing. This signs the right one.

## Root cause (from the rejected 26.2.20 IPA, build 669000012)

ITMS-91065 does **not** check the embedded framework's code signature. It checks each commonly-used third-party SDK's **xcframework origin signature**, which the archive records into the IPA at `Signatures/<name>.xcframework-ios.signature`. For sqflite that file said:

```
{ "signed" => false, "isSecureTimestamp" => false,
  "metadata" => { "library" => "sqflite_darwin.framework", "platform" => "ios" } }
```

`signed=false` → "Missing signature." We'd been signing the inner `.framework`, but the `.xcframework` **bundle** stayed unsigned. (And the host export re-signs the embedded slice at export time anyway, so the inner signing was a no-op that never reached the shipped binary — which is why #1133/#1135 didn't move the needle.)

This is exactly the feature behind the doc Apple's rejection links to ("Verifying the origin of your XCFrameworks"), and the same `SignatureCollection` step that originally crashed the archive.

## Fix

Apple's canonical command — and the sqflite maintainer's accepted fix ([tekartik/sqflite#1129](https://github.com/tekartik/sqflite/issues/1129)): sign the `.xcframework` **bundle** with `--timestamp`.

- **`sign-ios-frameworks.sh`**: sign the plugin `.xcframework` bundles with `codesign --force --timestamp --sign …` (was: inner per-slice `.framework`, no timestamp). Still excludes `Flutter` (already Google-signed; re-signing it crashed the archive in 26.2.18), `App`, and `FlutterPluginRegistrant` (not third-party SDKs). Verifies each carries a secure timestamp before the archive runs.
- **`verify-ios-ipa-signatures.sh`**: add the **real ITMS-91065 gate** — parse `Signatures/*.xcframework-ios.signature` and require `signed=true` + `isSecureTimestamp=true` for each non-excluded SDK xcframework. The embedded code-signature check stays as a defensive secondary check.

No Makefile change — same `sign-ios-frameworks` target and verify wiring from #1135.

## On reverting #1135 / #1136 (you asked)

Neither should be reverted:
- **#1136** (SIGPIPE fix) — pure correctness fix to the verifier we keep; reverting reintroduces `make Error 141`.
- **#1135** — keep the Flutter/App exclusion, the verify gate, and the Makefile wiring. Only its *inner-framework signing* was the wrong target; this PR replaces that in place. The exclusion is still load-bearing (signing Flutter's xcframework crashes the archive).

## Testing

- Verified against the **actual rejected 26.2.20 IPA**: the new gate FAILs it on all 9 plugin origin signatures (`signed=false`), and PASSes once the origin signatures are corrected.
- Confirmed locally that `codesign --force --timestamp --sign "Apple Distribution: …" sqflite_darwin.xcframework` yields `signed`=valid with a secure `Timestamp=` and passes `codesign --verify`.
- Selection logic signs the 9 plugin xcframeworks and skips Flutter/App/FlutterPluginRegistrant.
- Both scripts parse under stock `/bin/bash` 3.2.

## Confidence & residual risk

High confidence on the diagnosis (the `.signature` file literally says `signed=false`) and that signing the bundle with `--timestamp` flips it. The one thing only a real archive can confirm is that Xcode's `SignatureCollection` records `signed=true`/`isSecureTimestamp=true` from the signed bundle — **the verify gate now checks exactly that in CI and fails the build before upload if it didn't**, so the next release either ships clean or fails safely without burning an Apple round-trip.

## Operator action required

Merge, then cut a new iOS release tag (e.g. `26.2.21`). The build's verify step will confirm the origin signatures before upload.

### DIFF
--- a/scripts/sign-ios-frameworks.sh
+++ b/scripts/sign-ios-frameworks.sh
@@ -1,27 +1,33 @@
 #!/usr/bin/env bash
 #
-# Code-sign the add-to-app Flutter plugin xcframeworks before the host Xcode
-# archive embeds them.
+# Sign the add-to-app Flutter plugin XCFRAMEWORK BUNDLES (with a secure
+# timestamp) before the host Xcode archive consumes them.
 #
-# Why this exists:
+# Why this exists / what ITMS-91065 actually checks:
 #   `flutter build ios-framework` (see common/Makefile `build-ios`) emits every
-#   plugin as an UNSIGNED .xcframework. The host project embeds each with
-#   "Code Sign On Copy", but that runtime re-sign is unreliable for the device
-#   slice of a pre-built xcframework (flutter/flutter#148300, #179634), so the
-#   embedded framework can ship unsigned. App Store Connect then rejects the
-#   upload with ITMS-91065 ("Missing signature") for commonly-used SDKs such as
-#   sqflite (sqflite_darwin.framework), path_provider and shared_preferences.
+#   plugin as an UNSIGNED .xcframework. App Store Connect then rejects the upload
+#   with ITMS-91065 ("Missing signature") for commonly-used third-party SDKs such
+#   as sqflite (sqflite_darwin), path_provider and shared_preferences.
 #
-#   Signing the inner .framework of each slice here guarantees a valid signature
-#   is present regardless of whether Code-Sign-On-Copy runs: if Xcode re-signs on
-#   copy it harmlessly overwrites ours with the same Apple Distribution identity;
-#   if it skips the slice, our signature is what ships. Signing the .xcframework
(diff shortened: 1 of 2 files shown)


kar:
@codex review this

PR #1146, opened 2026-06-26
test(appium): paywall smoke scenario + automated account control

kar:
## What & why

Adds a second iOS Appium/WDIO smoke scenario: **open the Adapty paywall, screenshot it, and strict-compare to a committed golden**.

The paywall only appears for an **inactive (libre)** account — the opposite of the existing DNS-onboarding scenario, which needs an **active** account. Today the test device's account is pinned active by hand. So the core of this change is an **automated way to switch account state before each scenario**, replacing the manual pin.

## How account control works

It reuses the app's existing command bus through the support chat: any message prefixed `cc ` is routed to `command.onCommandString`, so `cc restore <id>` runs the `restore` command. This works on the real `six` debug build with **no app code change** and no gating.

- `flows/account.ts` — `ensureAccountActive()` / `ensureAccountInactive()` restore a dedicated account by id (`BLOKADA_ACTIVE_ACCOUNT_ID` / `BLOKADA_INACTIVE_ACCOUNT_ID`). Each spec self-provisions its state, so the suite is order-independent and self-healing. `restoreAccount` waits for a history-proof completion signal before relaunching to a clean Home.
- The power toggle's `value` reflects **protection on/off, not account state**, so account state is verified **behaviorally** (paywall appears for inactive; protection/DNS path for active), with negative guards that fail clearly if the account turns out wrong.

## What's included

- `specs/smoke/paywall.spec.ts` — restore inactive → tap power → assert native paywall presents → screenshot → strict golden compare.
- `support/golden.ts` — `pixelmatch` + `pngjs` comparison: status-bar masking, configurable threshold, `UPDATE_GOLDEN=1` to (re)generate, diff/actual artifacts on mismatch.
- `wdio.conf.ts` — explicit ordered `specs` (paywall first → suite ends with the device **active**, the default most scenarios expect). New specs must be registered here.
- `dns-onboarding.spec.ts` — added a self-healing `before` hook (`ensureAccountActive`) + relaunch to clear any leftover modal.
- `appium-smoke.yml` — passes the two account secrets to the run step.
- devDeps: `pixelmatch`, `pngjs`, `@types/pngjs`.

## Verification done

- `tsc --noEmit`: no errors in new/changed files (only pre-existing baseline noise).
- `make -C automation/appium test`: 100/100 harness unit tests pass.
- Golden mask + diff mechanics validated against synthetic PNGs.
- One code-review pass addressed (leftover-modal lifecycle, restore-race completion gate, strict-golden artifacts, false-pass guards). Not yet run end-to-end on a device.

## Operator actions required before this is green in CI

1. **Create two GitHub secrets**: `BLOKADA_ACTIVE_ACCOUNT_ID` (active, non-family) and `BLOKADA_INACTIVE_ACCOUNT_ID` (inactive/libre, non-freemium, non-family). Use **throwaway dev accounts** — the id is typed into the support chat and can appear in CI failure artifacts.
2. **Capture + commit the golden baseline** on the CI device: `UPDATE_GOLDEN=1 make appium-test`, then commit `automation/appium/wdio/src/specs/smoke/__golden__/paywall.png`. Strict compare fails until it exists. Regenerate after any intentional paywall change.
3. **Live-validate selectors** with the appium harness: home gear, support row, the unlabeled chat **send button** (coordinate-tap fallback — most fragile), and the Adapty paywall CTA predicate.

## Run locally before merge

```bash
export BLOKADA_ACTIVE_ACCOUNT_ID=... BLOKADA_INACTIVE_ACCOUNT_ID=... IOS_AUTO_SELECT_FIRST=1
make -C common build-ios
UPDATE_GOLDEN=1 make appium-test   # first time: writes the golden; inspect + commit it
make appium-test                   # strict compare, as CI runs it
```

### DIFF
--- a/.github/workflows/appium-smoke.yml
+++ b/.github/workflows/appium-smoke.yml
@@ -50,6 +50,11 @@ jobs:
         env:
           IOS_AUTO_SELECT_FIRST: "1"
           SHOW_XCODE_LOG: "0"
+          # Dedicated dev accounts the specs restore via the support-chat command
+          # bus: paywall needs inactive (libre, non-freemium, non-family),
+          # DNS onboarding needs active (non-family).
+          BLOKADA_ACTIVE_ACCOUNT_ID: ${{ secrets.BLOKADA_ACTIVE_ACCOUNT_ID }}
+          BLOKADA_INACTIVE_ACCOUNT_ID: ${{ secrets.BLOKADA_INACTIVE_ACCOUNT_ID }}
         run: make appium-test
 
       - name: Upload appium screenshots on failure
--- a/automation/appium/README.md
+++ b/automation/appium/README.md
@@ -290,7 +290,22 @@ If the session drops unexpectedly, first suspect device auto-lock or lost foregr
 
 ## Extending the Suite
 
-- Add specs under `automation/appium/wdio/src/specs/`.  
+- Add specs under `automation/appium/wdio/src/specs/` **and register them in the
+  explicit `specs` list in `wdio.conf.ts`** (the glob was replaced so spec order
+  is deterministic). Order by account state: inactive-account scenarios first,
+  active-account scenarios last, so the suite ends with the device active.
+- Control account state with `flows/account.ts`: `ensureAccountActive()` /
+  `ensureAccountInactive()` restore a dedicated account via the support-chat
+  command bus (`cc restore <id>`), reading the ids from the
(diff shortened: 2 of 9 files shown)


kar:
Superseded by #1147 — this PR auto-closed when its head branch was renamed to `appium-paywall-smoke`.

PR #1160, opened 2026-07-03
feat(ui): adaptive widescreen/tablet layout overhaul

kar:
## What

Overhauls the bolted-on widescreen/iPad UI into a proper adaptive layout system. Eight staged commits, each independently green (`fvm flutter analyze` errors-only gate + full test suite).

**User-visible on iPad / wide windows:**
- No more blank right pane: Activity and Privacy Pulse use the supporting-pane pattern — with nothing selected the master renders centered at its comfortable single-pane width (no empty half at all); selecting a domain animates the master aside while the detail pane slides in from the right (250ms). Settings and family Device keep their natural default detail and open split from the start.
- Content everywhere gets a consistent 16pt breathing gap below the glass top bar (it used to sit flush against the bar's bottom edge on iPad; only Exceptions hand-added spacing).
- The freemium gate centers on the viewport with a full-body blur (was centered inside the left pane only), while the top bar and back stay usable.
- Two-pane layouts start at ≥ 840pt width with a ≥ 500pt height guard (was > 1000pt): iPad 2/3 Split View and 11" landscape splits gain master-detail; landscape phones stay single-pane.
- Live re-layout on window resize/rotation with pane selection preserved (the old mode was cached at app start, so taps could route to a stale pane after resize).
- Tablet Settings gains the previously missing Add-exception top-bar action (with its `automation.exception_add_button` Semantics); Support always opens full-screen (its pane never worked); opening a detail with no pane visible falls back to a normal push (was a stale-closure no-op / blank pane).

**Phone rendering is intentionally unchanged.** No visual re-skin.

## How

New `common/lib/src/shared/layout/`:
- `window_shape.dart` — reactive per-build window classification (compact/medium/expanded) + breakpoint constants; single home for `maxContentWidthTwoPane`, `minWidthFloatingSheet`.
- `detail_route.dart` — `DetailRoutes` registry: each route's title/body/pane-body/trailing defined once, consumed by `generateRoute`, `TopBarController.didPush` and the detail pane (kills the 3× duplication that had already drifted).
- `detail_pane_host.dart` — scoped pane hosts replacing the global last-`initState`-wins `Navigation.openInTablet` static; hosts are gated on `mounted && ModalRoute.isCurrent && expanded`.
- `with_detail_pane.dart` + `detail_pane_placeholder.dart` — the reusable master-detail scaffold (one `WithTopBar` in both modes → exactly one `screenTitle` node) used by Settings, family Device, v6 Activity and Privacy Pulse.
- `WithTopBar` gains an `overlay` slot (above content, below top-bar chrome) hosting the freemium gate.

Deleted: `Navigation.openInTablet`, `Navigation.isTabletMode` (+ its `app.dart` write), `Paths.openInTablet` flag, `isTabletMode(context)`, `maxContentWidthTablet`, `V6FiltersSection.twoColumns`, dead `primary:`/`freemium` section params, dead `Paths.settingsAccount`.

`Navigation.open` signature unchanged; all call sites untouched.

## Verification

- 407 tests green (baseline 379 + 28 new: breakpoint matrix, registry title/args parity pins for both flavors, pane host lifecycle incl. real-Navigator `isCurrent` gating, resize round-trip, overlay geometry).
- iPad Pro 13" mocked simulator, driven via the Appium explorer: Settings two-pane + pane swap + in-pane toggle; Privacy Pulse master + placeholder → domain tap fills pane; Advanced two-column filters; exceptions-pane row → full-screen push fallback → back returns to intact two-pane; `automation.exception_add_button` present on tablet.
- Two review passes by independent reviewers; all findings fixed in the final commit (restored `activitySearch` Semantics, autorun disposers, placeholder guard, pop-hazard doc, isCurrent lock-in test).

## Not covered here (follow-ups / operator)

- Freemium gate visual on device needs a freemium-flagged (inactive) account — the seeded dev account is active cloud. The overlay geometry is widget-tested; a real-device freemium CTA → Adapty paywall smoke is required before release.
- `DetailPanePlaceholder` remains only as a fallback for paths without a registered pane body (the supporting-pane pattern removed its normal appearance); hint copy via the translate submodule is optional now.
- Product decision: pane vs pushed rendering of journal entries still diverges (`StatsDetailSection` vs `DomainDetailSection`) — preserved as-is, consolidation tracked separately.
- Family deep-polish (per agreed scope: v6 first); family Device screen migrated and unit-tested. Note: FamilyMocked on the iPad Pro (M5) simulator hangs on a white screen before Flutter's first frame **on `main` too** (verified by differential build on a clean sim) — a pre-existing mocked-sim issue, not from this branch; family on-device verification blocked on that.
- Android tablet emulator pass and Mac (Designed for iPad) resize pass.


kar:
## Screen-by-screen refinement round (commits `f190d6cb`..`1ff4ff60`)

Iterated on the wide-screen layouts with live on-device (simulator) feedback after the initial overhaul landed:

**Global**
- Fixed content sitting flush against the glass top bar: the 16pt breathing gap now lives in `getTopPadding` itself, so every screen inherits it (Exceptions was the only screen hand-adding it before).
- Supporting-pane pattern in `WithDetailPane`: hosts choose between master-solo-until-selected (Privacy Pulse) and permanent split with a designed placeholder (Activity, Settings). Split widths animate from a single tweened progress value derived from the live box each frame — window resizes mid-animation can't overflow, and narrow boxes (portrait iPads) skip the split entirely and push details full-screen (`minSplitMasterWidth`) instead of tearing through a mid-flight column collapse.
- Unified list-selection style across all selectable rows (top domains, recent activity, settings items, journal entries): full-row square tint in a shared accent-derived color (`theme.selection`), with the open detail's row highlighted via the new `PaneSelection` inherited widget — scoped to the selection's source so a toplist selection doesn't light up a same-named recent-activity row (and vice versa).

**Privacy Pulse**
- Solo (nothing selected): full-width two-column dashboard — chart + recent activity left, all-time counter + top domains right, one shared scroll.
- Selecting a domain animates to a 60/40 split; the master keeps both columns (they shrink to fit alongside the pane) and the detail gets its designed 500pt.

**Settings**
- No more silently-preselected Exceptions: opens 50/50 with a "Select an item to see details" placeholder; the selected row highlights.
- "Chat with us" now opens in the pane like the other rows — End-chat only pops when Support was pushed, which was the reason it had to be excluded before.
- The previously missing Add-exception top-bar action appears whenever Exceptions is shown.

**Activity (See All)**
- Same placeholder-split treatment (the solo centered journal read as empty); tapped journal entries highlight by identity (domain + timestamp).
- Pushed domain-details pages get a 700pt content column on wide windows (was 500).

Follow-ups unchanged: placeholder copy needs a translate-submodule key (English literal for now), real-device freemium smoke, Android tablet / Mac resize passes.

kar:
## Smoke failure debugged: pre-existing device flake, not this branch — deflaked in `86bffa00`

**Symptom** (run 28678062001): only `activity.spec.ts` fails — after tapping Privacy Pulse's "See All", `~automation.screen_activity` never appears. All 10 other specs pass on this branch, including the ones covering the new adaptive layouts (settings-navigation, exceptions-crud, privacy-pulse-range, tab-navigation).

**Root cause:** not a regression from this PR.
- The identical failure hit **`main`** at 13:51 and 14:22 today (runs 28664881239, 28666499717) — after the same commit had **passed** at 12:34 and 13:11. Same code, both outcomes → device-side flake, started mid-day independent of any change.
- The failure artifacts show the mechanism: the "See All" button is present, visible and correctly positioned in the accessibility tree, WDA's `element.click()` reports success — but the failure screenshot still shows Privacy Pulse and the app's own log has no tap trace. WDA's click on this Flutter button intermittently never reaches the `GestureDetector` (a known Flutter/XCUITest quirk we already work around elsewhere), with the scroll-settle race after `scrollToShowAll` widening the window as the device's real journal grows.

**Fix** (spec-only): after scrolling to "See All", poll until its position is stable, dispatch a real coordinate tap (`mobile: tap` at the freshly-resolved center — same primitive `account.ts`/`onboarding.ts` already use on this device), and retry up to 3× before the authoritative `waitForScreen` assertion. Verified `tsc`-clean against the harness baseline; re-running the smoke workflow on this PR to confirm on the device.

Note: `main`'s smoke will stay red on this spec until this lands.

kar:
Smoke re-run on this branch is green: 11/11 specs passing on the device, including the deflaked `activity.spec.ts` (run 28680053212). https://github.com/blokadaorg/blokada/actions/runs/28680053212

kar:
## Advanced screen: flowing, height-balanced blocklist columns (`9a5b0bc6`, `199ebd48`)

The wide two-column blocklist layout had two problems, both fixed and verified on the iPad simulator:

- **Dead space between rows** — cards were paired into `Row`s, so every pair stretched to its taller card's height (blocklist cards vary a lot: Steven Black has 5 options, AdAway 1). The cards now flow in two independent `Column`s that each stack tightly.
- **Lopsided columns** — plain index alternation piled the tall cards onto one side, ending the right column two cards early. Distribution is now greedy by estimated card height (option rows dominate at ~60pt each, plus wrapped description lines), so both columns end together. The estimate only needs relative accuracy; if a future card set defeats it, the upgrade path is measured masonry.

Bonus fix the wide cards exposed: `FilterWidget`'s outer `Column` defaulted to center alignment, so a card whose description fits on one line (Dan Pollock's) had its whole header block centered — reading as a stray left indent. The column now stretches its children; phone rendering is unaffected (descriptions always fill the narrow column there).

Phone layout untouched throughout (single column as before). Suite green, analyze gate clean.

PR #1174, opened 2026-07-13
chore(deps): upgrade Flutter 3.35.2 → 3.44.6, vendor input_code_field, drop the adapty fork

kar:
Follows #1172 and #1173 (both merged).

## Why

Flutter **3.35.2** (Aug 2025, Dart 3.9.0) was ~11 months and ~66 stable releases behind. It also silently blocked `flex_color_scheme` **8.4.0** and `flex_seed_scheme` **4.x** — both need Flutter ≥3.38 — which are exactly the theming packages Dependabot kept trying and failing to land in #1149. Now on **3.44.6 / Dart 3.12.2**.

## The only real blocker was one abandoned package

`input_code_field` **2.0.2** is the last release (2023-11-11; two releases ever). Flutter 3.44 added `onFocusReceived` to `TextInputClient` — it *has* a default body, but `_InputCodeFieldState` uses `implements`, which inherits no implementations, so the class must declare the member itself, and upstream never will. That one missing override failed **all 50** of the compiling tests.

**Vendored** to `common/lib/src/shared/ui/vendor/input_code_field/` (MIT; LICENCE and attribution preserved), one method added, pub dependency dropped. Used in exactly one place — the app-lock PIN field in settings.

> Worth recording: the **pigeon-12 / analyzer-5.13 hold was NOT a blocker.** Resolution, `make gen` (249 outputs, 1706 actions) and `analyze` all pass on Dart 3.12 with pigeon held exactly as-is. A Flutter upgrade does **not** require scheduling the pigeon 26 migration.

## Two build breaks, both fixed

**iOS.** Flutter 3.44 added native XCFramework code-signing to `build ios-framework` and defaults it **on**. It fails on any machine holding two identities for one team (`Multiple identities found for development team …`) — which includes the CI mac runner. We already sign these ourselves in `sign-ios-frameworks` (the ITMS-91065 fix), so `--no-codesign` restores the pre-3.44 behaviour and leaves that pipeline untouched. Flutter's own `--codesign-identity` is a candidate to *replace* that script later; not here.

**Android.** Flutter 3.44's module template moved to **Gradle 9**, which refuses adapty's versionless Compose transitive (`Could not find androidx.compose.ui:ui:.`). Fixing that surfaced a second issue — the plugin compiles against `android-34`, but Compose `material3 1.4.0` needs ≥35.

## …which let us delete the adapty fork entirely

The fix was **not** to patch our fork. `adapty_flutter` **3.17.1** (2026-07-02) ships:

> - [Android] Updated native Android SDK to 3.17.2.
> - **[Android] Fixed AGP 9 and Built-In Kotlin support.**

…plus `jvmTarget = JvmTarget.JVM_11` — **the entire reason the `blokadaorg/AdaptySDK-Flutter` fork existed.** Verified: plain upstream 3.17.1 builds the AAR under Gradle 9, which the old fork (based on 3.17.0) does not. So:

- `adapty_flutter: 3.17.0` + git-SHA `dependency_overrides` → **`adapty_flutter: 3.17.1` from pub.dev**. Fork dropped, override block deleted.
- Bonus: dep-validate's `queue.mjs` can now see adapty as a **payment** surface again, instead of an opaque git SHA (a known gap).

## Verification — full local matrix on 3.44.6

| check | result |
|---|---|
| `make test` | **379/379 pass** |
| `flutter analyze --no-fatal-warnings --no-fatal-infos` | exit 0 |
| `make pub-verify` (the lock gate from #1173) | exit 0 |
| `make -C common build-ios` | ✅ |
| iOS `Dev` debug / `FamilyDev` debug | **BUILD SUCCEEDED** ×2 |
| `make regen-android` (AAR) | ✅ |
| `apk-six-debug` / `apk-family-debug` | **BUILD SUCCESSFUL** ×2 |

`flex_color_scheme` deliberately stays at **8.3.1** — `pub get` honours the lock, so this is an SDK upgrade, not a dep-bump spree. Dependabot can now offer 8.4.0 *for real*.

## ⚠️ Needs on-device sign-off before merge

**adapty moved in this PR** (3.17.0 + fork → 3.17.1 upstream), and adapty is the payment surface. Unit tests and builds cannot validate it. The Appium smoke runs on the physical device for this PR and covers paywall display — but **purchase and restore go through the StoreKit sandbox sheet, which no agent can drive.** Those need a human confirmation on the device, or an explicit acceptance of the residual risk.

Also note the 9-minor Material/theme jump: widget tests pass, but that is not the same as eyes on the UI.


kar:
## Device smoke: green (11/11) — after it caught a blank-screen regression

The **first** smoke run on this branch [failed every spec](https://github.com/blokadaorg/blokada/actions/runs/29234615191), and it was right to. The app came up on a **blank white screen** on both the simulator and the physical device, while CI was fully green (both platforms built, 379/379 tests, analyze clean). Root cause:

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception:
Unable to load asset: "AssetManifest.json".
```

`i18n_extension_importer`'s `JSONImporter.fromAssetDirectory` reads the legacy `AssetManifest.json`, which Flutter replaced with `AssetManifest.bin` and no longer ships. The throw escaped `I18nService.loadTranslations()` at startup, nothing caught it, and the widget tree never mounted — so every spec died in its `before all` hook with no automation ids, because there was no UI.

That package is **abandoned too** (0.0.6 is the latest, 2023-11-13) — the second one in this PR, after `input_code_field`. No version to bump to. It was used for one method across three call sites, so the dependency is **dropped** and the loader inlined on Flutter's supported `AssetManifest.loadFromAssetBundle` API.

Also fixed: `--no-codesign` had been applied to only **1 of 5** `flutter build ios-framework` invocations. The other four — including `common/scripts/build.for.ios.sh`, the **release** path — would still have failed on a multi-identity machine.

### Now green on `f939dba`

CI 5/5, and the [physical-device smoke](https://github.com/blokadaorg/blokada/actions/runs/29237069740) passes **11/11**: home hub, power pause / turn-off, protection persists across cold restart, DNS onboarding, **paywall on inactive account**, account subscription status, blocklist pack toggle, custom exception CRUD, Activity, Privacy Pulse, Settings sub-pages.

### Still outstanding before merge

- **Purchase + restore are NOT covered.** The StoreKit sandbox sheet is a system UI no agent can drive, and **adapty moved in this PR** (3.17.0 + fork → 3.17.1 upstream). The paywall *renders*, but a real purchase/restore needs a human on the device — or an explicit acceptance of the residual risk.
- **Two abandoned packages have now broken on 3.44 for unrelated reasons.** A manual pass over the app is worth the time before merging; there may be a third.

kar:
Tested manually on device. Looks fine.

PR #1182, opened 2026-07-20
fix(appium): fail fast on blocking system alerts + capture hook-failure artifacts

kar:
Two independent automation fixes, one commit each — splittable if you would rather land them separately.

## 1. Appium: diagnose blocked-device runs (`4680804`)

An **"Apple Account Sign In Requested"** 2FA modal sat on the CI iPhone for four consecutive `appium-smoke` runs (2026-07-20, runs `29714640386` → `29720164508`). It is a SpringBoard alert, so it swallows every tap while the Flutter view stays in the accessibility tree — the gear was *found* and *clicked* successfully, the taps just never reached the app. All 11 specs then failed in `before all` on an unrelated selector:

```
element ("~automation.screen_settings") still not existing after 15000ms
  at openSettingsScreen (src/flows/account.ts:64)
```

Confirmed present 20–22× in every red run and **absent from the last green run**.

- **Fail fast.** `acceptNotificationAlert` already detected the alert and had its text, but returned silently to "leave it for the caller to handle" — and no caller anywhere does. It now throws `BlockingSystemAlertError` with the text for device-level modals, and still returns silently for flow alerts.
- **Narrow on purpose.** A missed pattern falls back to today’s timeout; a false positive fails a run that would have passed. It must **not** match `Apple ID`/`Apple Account` alone — the StoreKit sandbox purchase sheet prompts for Apple ID sign-in mid-flow (dep-validate Stage D drives purchase/restore) — so only the 2FA phrasing matches.
- **Artifacts on hook failure.** This needed server-log forensics because the run produced no visual evidence: spec screenshots are per-test, and a `before all` failure aborts before any test body runs (the job reported *"No files were found with the provided path: output/*.png"*). `afterHook` now captures a screenshot + page source, best-effort so it cannot mask the hook’s real error. Both globs are already uploaded by the workflow.

## 2. dep-validate: drop the retired github-actions carve-out (`426dfd4`)

`dependabot-auto-merge.yml` removed its github-actions branch-prefix case in #1115, but `queue.mjs` / `SKILL.md` still ported the old rule — so every action bump was queued as needing human validation at any bump level, while the workflow auto-merged the non-major ones.

## Verification

- `tsc --noEmit` diffed against a stashed baseline: **zero new errors** (the 5 reported pre-exist).
- `npm run test:unit`: **93/93** (was 91).
- New test parses the regexes out of `alerts.ts` rather than copying them, so it cannot drift. Confirmed it **goes red** when the over-broad `/apple id/i` pattern is reintroduced.
- Queue: #1177/#1179 still queue via major + high-risk; a non-major action bump now yields zero reasons.

**Not verified on-device.** `afterHook` firing and the error surfacing through the reporter are only provable on a real run. Smoke dispatched against this branch to check exactly that — the device still has the modal, so the expected result is a **fast, clearly-labelled failure with a screenshot**, not a green run.

### DIFF
--- a/.agents/skills/dep-validate/SKILL.md
+++ b/.agents/skills/dep-validate/SKILL.md
@@ -21,7 +21,7 @@ node automation/dep-validate/queue.mjs
 
 Prints JSON: `queue` (open Dependabot PRs needing validation) and `advisories` (unscanned ecosystems). Each PR entry carries `reasons`, `majors`, `highRiskPackages`, and `surfaces` (e.g. `payment`, `platform-bridge`, `protection`, `messaging`, `build-system`) precomputed from the bump. Validate one PR per run; take the PR number as input for headless runs.
 
-A queue entry means: the auto-merge workflow would hand it to a human (major bump, non-allowlist file, or github-actions), or it carries a high-risk package (Adapty, Firebase, WireGuard, pigeon, gradle/AGP) at any bump level. Advisories are version-drift notes only — there is no PR to check out, so you surface them, you do not run the loop on them.
+A queue entry means: the auto-merge workflow would hand it to a human (major bump or non-allowlist file), or it carries a high-risk package (Adapty, Firebase, WireGuard, pigeon, gradle/AGP) at any bump level. Advisories are version-drift notes only — there is no PR to check out, so you surface them, you do not run the loop on them.
 
 ## 2. The validation loop
 
--- a/automation/appium/wdio/scripts/tests/blocking-alert-patterns.test.mjs
+++ b/automation/appium/wdio/scripts/tests/blocking-alert-patterns.test.mjs
@@ -1,0 +1,63 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+// alerts.ts is TypeScript and imports @wdio/globals (needs a live session), so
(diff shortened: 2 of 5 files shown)


kar:
On-device verification — smoke dispatched against this branch: [run 29726935031](https://github.com/blokadaorg/blokada/actions/runs/29726935031). Still red (the modal is still on the device), which is the expected and correct outcome. What changed is the diagnosis:

**Before**
```
element ("~automation.screen_settings") still not existing after 15000ms
  at openSettingsScreen (src/flows/account.ts:64)
```

**Now**
```
BlockingSystemAlertError: Unexpected system alert is blocking the device:
"Apple Account Sign In Requested
Your Apple Account is being used to sign in on the web.".
It swallows every tap, so the app cannot be driven. Dismiss it on the device and re-run.
```

- Old misleading error: **0 occurrences** in the new run.
- `appium-screenshots-29726935031` uploaded (417 KB) — previously the step reported *"No files were found"*. Each `hook-failure-*.png` shows the modal with its Don't Allow / Allow buttons; `.xml` page source captured alongside.
- Faster failure: specs **2:40 vs 6:21**, job **7m46s vs 11m22s**, since each spec throws on detection instead of burning two 15s timeouts.

Nit for a follow-up, not fixing here: the generated filenames are long and truncate mid-word (`...before-all-hook-wrappedhook-for-renders-.png`) because mocha's hook title is `"before all" hook for "<test>"`. Functional, just ugly.

PR #1196, opened 2026-07-28
feat(release): continuous delivery pipeline rework

kar:
Part of blokadaorg/issue-tracker#324. Design doc and runbook: [`docs/release-pipeline.md`](docs/release-pipeline.md).

Supersedes #1195 (GitHub closed it irrecoverably during a branch rename).

## What changes

- **Every merge to `main`** allocates a build number, builds all four artifacts (six + family × iOS + Android), and uploads to Play **internal** and **TestFlight**. Nothing published, nothing submitted.
- **Releasing never builds.** `release.yml` is dispatch-only and promotes an already-uploaded build to Play `alpha`+`beta` and to an App Store version, optionally submitted for review. Publishing stays a manual click.
- **Versions are `YY.M.BUILD`** (e.g. `26.7.1000`); store codes are `669000000 + BUILD`, allocated by an annotated `build/N` git tag.
- **`release-tag.yml` and OneDev are retired.**

The point: the binary you ship is the exact one that sat on internal, not a fresh build from the same commit.

## Live validation

The branch was pushed with a temporary branch-scoped trigger and `ci-release.yml` ran **green end to end** — all six jobs, `build/1000` allocated and annotated `26.7.1000`, all four artifacts uploaded to Play internal and TestFlight. The trigger was then removed (`958d8f98`); re-pushing after that fired nothing, confirming the removal.

**The promote path has never run against a real store.** It is dispatch-only and cannot execute until it is on the default branch.

## Review found five defects in the plan, not the implementation

Each task passed its own scoped review; a whole-branch review then caught what per-task gates structurally cannot. All five were silent-failure shaped — none would have reddened CI:

1. **OneDev's `publish android manually`** called `publish-android`, whose meaning changed — it would have kept reporting success while shipping to internal only, never reaching alpha/beta.
2. **Deleting `release-tag.yml` orphaned `translation_coverage.py`**, dropping per-locale reporting with no error. Now runs per merge in `ci-release.yml`.
3. **`release.yml` interpolated dispatch inputs into `run:` bodies** — script injection on a persistent self-hosted runner holding signing keys. The workflow it replaces already used `env:` indirection; the plan regressed it.
4. **Android builds were promotable for exactly one merge.** `supply`'s `update_track` *replaces* the internal track's release list, so build N vanished when N+1 landed. Now writes the target track directly via `--version_codes_to_retain`.
5. **The iOS promote never attached the binary** on the default path — `deliver` only consumes `build_number` inside its submit flow, so a release went green having created a version with metadata and no build. Now attached explicitly via `Spaceship::ConnectAPI`.

Also found along the way: `submit_for_review` was **unreachable** — fastlane coerces the CLI string to a Boolean, so `== "true"` was never true.

## Known-unproven on first use

- The new Play promote call and the iOS attach lane have never run against a real store.
- `submit_ios_for_review: true` has never executed (it was impossible until `306b9a51`).
- ~~Three parked Low findings~~ — all fixed in `565ffc4b` (see the follow-up comment).

## Follow-up cleanup

- The three parked Low findings are fixed (`565ffc4b`): the preflight's `LoadError` hint now names the right interpreter, its rescue is split so auth / edit-open / listing each report themselves, and the concurrency note covers both eviction directions.
- Dead Make targets removed now that the OneDev buildspec is gone: `ci-test` (`565ffc4b`) and `ci-copy-source` with its `CI_BUILD_DIR` (`ff9910dc`). Neither had a caller.
- The runner pre-check passed — fastlane **2.235.0**. Every load-bearing fastlane behaviour was re-verified against 2.235.0 source rather than the local 2.232.0; nothing that matters changed. Details in the comments.

## Operator action required

**Before the first `make promote-android`** (not before merge), on the self-hosted runner:

```
ruby -e "require 'supply'; puts Fastlane::VERSION"
```

The preflight runs under `#!/usr/bin/env ruby`, while `$(FASTLANE)` is Homebrew's wrapper with a private `GEM_HOME`. If this fails, install the `fastlane` gem for the `ruby` on that runner's PATH — the Homebrew formula does not satisfy it. It fails closed, so the only cost is a blocked release, but it blocks every Android release until resolved.

**After merge**, the next push to `main` triggers a real four-artifact build and store upload — there is no dry-run mode. Expect `build/1001`.


kar:
## Status: complete and ready for review

27 commits, `main` is an ancestor (0 behind / 27 ahead), CI green, whole-branch review verdict **Ready to merge — yes**.

All eight planned tasks landed, each gated by its own spec+quality review, followed by a whole-branch review and two fix rounds. Every task's review ran in a context that didn't write the code.

## What has actually been tested

Three tiers, and the distinction matters — a lot of this code can only be proven by running it against a real store.

### Proven live against real infrastructure

- **`ci-release.yml` ran end to end, green.** Run [30357266362](https://github.com/blokadaorg/blokada/actions/runs/30357266362): all six jobs, ~51 min wall-clock. It allocated `build/1000` annotated `26.7.1000`, built all four artifacts, and uploaded them to **Play internal** and **TestFlight** for both flavors. The store code was `669001000`, above the previous max of `669000015`.
- **The temporary branch trigger was removed and verified.** Re-pushing after `958d8f98` fired no run, confirming the workflow now triggers on `main` only.

### Automated tests

| Suite | Result |
|---|---|
| `make test` (Dart) | 379 tests, all passed |
| `make test-scripts` | 2/2 pass |
| PR CI | 5/5 — en.json drift, Flutter analyze, Android assemble, iOS build, **Release script tests** |

`make test-scripts` is new and now runs in PR CI on `ubuntu-latest`. It guards the two invariants that can't be recovered from once broken: monotonic store codes, and no double-applied offset. `test-build-number.sh` covers the retry/lock path by installing a real `pre-receive` hook so git itself rejects the push — the retry branch is exercised through the actual protocol, not mocked — plus credential redaction and target validation.

### Verified by source reading only — NOT executed

These were checked against the installed fastlane 2.232.0 source rather than documentation, but have never run:

- **The Play promote call.** `--track <alpha|beta> --version_codes_to_retain N` writing the target track directly. The riskiest unproven assumption in the branch is that Play accepts `edits.tracks.update` naming a version code that isn't on any track.
- **The iOS `attach_build` lane** (`Spaceship::ConnectAPI`).
- **The Ruby preflight's success path** — exercised against a stubbed client and a live auth rejection, never real Play credentials.
- **`submit_ios_for_review: true` has never executed at all.** It was literally unreachable until `306b9a51`: fastlane coerces the CLI string to a Boolean, so the lane's `== "true"` comparison was always false. Enabling it will do something this codebase has never done.

## Next steps

**Before merge:** nothing. The branch is green and mergeable.

**On merge — this is the one to plan for.** The next push to `main` triggers a real four-artifact build and store upload. There is no dry-run mode and no path filters. Expect `build/1001`. Make sure the self-hosted pool and the Play/ASC credentials are healthy first.

**Before the first `make promote-android`** (not before merge), on the `[self-hosted, macOS, ARM64, build-only]` runner:

```bash
ruby -e "require 'supply'; puts Fastlane::VERSION"
```

The preflight runs under `#!/usr/bin/env ruby`, while `$(FASTLANE)` is Homebrew's wrapper with a private `GEM_HOME` — the Homebrew formula does **not** satisfy this. It fails closed, so the only cost is a blocked release, but it blocks *every* Android release until fixed.

**Suggested first exercise of the promote path:** dispatch `release.yml` narrowed — `platform: android`, `flavor: six`, `submit_ios_for_review: false`. A narrowed run produces no record tag by design, which makes it a safe way to prove the promote call against the real store before a full release.

**Reaching production is still a console action.** The pipeline stops at alpha/beta; nothing here promotes to production.

## Parked findings

Three Low, all wording, none able to ship a wrong artifact:

1. `scripts/verify-play-version-code.rb:28` — the `LoadError` hint names the wrong fastlane install, and it fires in exactly the scenario flagged above.
2. `scripts/verify-play-version-code.rb:89-90` — the outer rescue prefixes auth failures as "could not list uploaded artifacts". The underlying Google message is verbatim, so it stays actionable.
3. `docs/release-pipeline.md:157-161` — describes a merge's queued run being evicted, but a pending *release* run can equally be evicted by a landing merge.

Happy to fix these in a follow-up or here, whichever you prefer.

## Also worth knowing

- `make ci-test` is now called by nothing — the OneDev buildspec was its only caller. Removing the target is a surface decision I left to you.
- `fastlane match` has no automated entry point any more. It never really did: the OneDev job had no `triggers:` block, and `AGENTS.md` already documents it as a local step.
- With both workflows sharing the `store-publish` concurrency group, two merges landing during a release will evict the earlier queued run, so that commit is never built. This partially undercuts the "every commit on main is proven publishable" property and is documented as such.

kar:
## Operator pre-check passed — and a correction to the verification basis

`ruby -e "require 'supply'; puts Fastlane::VERSION"` on the runner returns **2.235.0**. The preflight's interpreter can load supply, so the blocking operator action from the previous comment is cleared: `make promote-android` will not fail closed at the preflight.

That surfaced something worth recording. **Every fastlane behaviour in this branch was verified against 2.232.0 (the local install); the runner has 2.235.0.** I've now diffed 2.235.0's actual source against 2.232.0 for every file the design depends on. Nothing that matters changed:

| Behaviour the design rests on | 2.235.0 |
|---|---|
| `update_track` replaces `track.releases` | unchanged (`uploader.rb:450`) — the Android promote defect is real on 2.235.0 too |
| `version_codes_to_retain` concat in `perform_upload` | unchanged (`uploader.rb:22`) — the fix's mechanism exists |
| `rescue_changes_not_sent_for_review` default | still `true` — passing `false` is still required |
| `aab_version_codes` / `apks_version_codes` / `make_from_config` | all present — the Ruby preflight works |
| `submit_for_review if options[:submit_for_review]` | unchanged (`runner.rb:74`) — the explicit attach is still necessary |
| `submit_for_review` / `skip_binary_upload` defaults | still `false` / `false` |
| `pilot` lane params (all 5) | all present |
| `fastlane/runner.rb`, `command_line_handler.rb`, `deliver/submit_for_review.rb` | **byte-identical** |

The last row covers the two subtlest conclusions in this work — the lane-body vs action working directory (which caused the iOS key-path Critical) and the `"true"`-to-Boolean coercion (which made `submit_for_review` unreachable). Both rest on files that did not change.

Non-behavioural diffs: supply's `options.rb` changed only `json_key` description text, and `uploader.rb`'s only change is a defensive `track.releases || []` nil-guard.

## One residual unknown on 2.235.0

Deliver 2.235.0 added two **unconditional** steps to its metadata upload path that 2.232.0 does not have:

```ruby
UploadAppClipDefaultExperienceMetadata.new.upload_metadata(options)
UploadAppClipDefaultExperienceHeaderImages.new.find_and_upload(options)
```

They run on every `deliver` call, which includes `promote-ios`. They are very likely no-ops for an app without an App Clip, but that is unverified — it cannot be confirmed without executing against App Store Connect. **If the first iOS promote fails inside metadata upload, look here first.**

The other 2.235.0 deliver changes (app-preview sync, the transporter Individual-API-Key rework) do not touch this path: app-preview sync is guarded by `options[:app_previews_path]`, which these lanes never set, and `skip_binary_upload: true` means no transporter runs at all.

## Revised next steps

1. ~~Runner pre-check for the preflight interpreter~~ — done, 2.235.0.
2. Merge when ready. The next push to `main` still triggers a real four-artifact build and store upload; expect `build/1001`.
3. First exercise of the promote path: dispatch `release.yml` narrowed to `platform: android`, `flavor: six`, `submit_ios_for_review: false`. Narrowed runs produce no record tag by design.
4. First iOS promote is the one to watch, for the App Clip step above.

kar:
## Addressed: the three parked findings, plus dead-target cleanup

Two commits: `565ffc4b` (the findings + `ci-test`) and `ff9910dc` (`ci-copy-source`).

### 1. `verify-play-version-code.rb` — the `LoadError` hint named the wrong install

It said *"This script needs the same fastlane install `make promote-android` uses."* That is actively misleading: `$(FASTLANE)` is Homebrew's wrapper with a private `GEM_HOME`, whereas this script runs under `ruby` from `PATH` and needs the gem visible *there*. Repairing or reinstalling the Homebrew formula — the obvious reading — would not have fixed it. The hint now says so explicitly and gives the actionable command.

This is the message that fires in precisely the scenario flagged as the top risk of the whole preflight, so a wrong hint had real cost.

### 2. `verify-play-version-code.rb` — the rescue reported the wrong stage

One `rescue` wrapped client construction, edit-open and listing together, so an `invalid_grant` came out as *"could not list uploaded artifacts"*. Split into three stages, each naming itself. Verified with a stubbed client, no network:

```
auth: Error: could not authenticate with Play for org.blokada.sex: invalid_grant: account not found
edit: Error: could not open a Play edit for org.blokada.sex: editAlreadyExists
list: Error: could not list uploaded artifacts for org.blokada.sex: permissionDenied
```

The throwaway-edit cleanup still runs when listing fails — `ensure` executes before `SystemExit` propagates — so no dangling Play edit, which was the behaviour worth preserving through the restructure.

### 3. `docs/release-pipeline.md` — the concurrency note described only half the eviction

It covered a merge's queued run being cancelled, but a pending **release** run is evicted the same way when a merge lands after the dispatch. That is the direction people actually notice, because the release disappears from the queue rather than failing. Both directions are now documented, along with the fact that nothing is left half-done (the release is promote-only) and it simply has to be dispatched again.

### 4. Dead Make targets

Both were OneDev leftovers with no caller anywhere:

- `ci-test` — the buildspec was its only caller, and it went with the buildspec.
- `ci-copy-source` — orphaned even before this branch. `CI_BUILD_DIR` existed only to feed it, so it went too.

## Verification

- `ruby -c scripts/verify-play-version-code.rb` — Syntax OK
- All three rescue stages exercised against a stubbed client (output above)
- `make test-scripts` — 2/2 pass
- `make -n` resolves for `test`, `test-scripts`, `publish-android`, `promote-android`, `publish-ios-testflight`, `promote-ios`, `version`, `ci-build-android-six` — the `.PHONY` list survived both removals with no dangling continuation
- `rg` confirms no reference to `ci-test`, `ci-copy-source` or `CI_BUILD_DIR` remains
- Pushing fired no `ci-release` run, as expected on a non-`main` branch

## Unchanged

Nothing here touches release behaviour — the fixes are two error strings, one rescue split, one doc paragraph, and two unused targets. The first-execution unknowns from the previous comments still stand: the Play promote call, the iOS attach lane, the preflight's success path against real credentials, `submit_ios_for_review: true`, and the two new unconditional App Clip steps in deliver 2.235.0's metadata path.

Issue #427, opened 2019-07-09
I want a smaller notification


depuytnl:
The new Blokada 4 notification is nice, but I don't mind too much what domains we blocked recently and would prefer for it to be much smaller, still allowing me to switch blokada on and off quickly, while not taking too much space; Can this be made a configuration option at least? (In my mind most users will not require the level of detail it currently gives, so this could be a good default, but opinions may differ!)

Cheers!

depuytnl:
found the option :O

pm-at:
Where is that option?

depuytnl:
What I thought I found seems to be a standard option for any notification to collapse it. Sadly it doesn't stay collapsed, so I disallowed notifications for the app altogether. It's a bit crude and makes it harder to toggle if blokada is enabled, but it solves the issue for me.

Issue #341, opened 2019-03-07
Why white-listing is required for some applications?


renaudcerrato:
Android Developer here. I'm a user of Blokada (Adblock flavor), and I'd like to contribute to it. I went through the source code, and I can't understand WHY some applications (like Google Play) needs to be white-listed in order to work, **even if Blokada is patched to not filter anything (forwarding only)**.

The VPN is advertising itself as a DNS server (on address xxx.xxx.xxx.2), and ONLY route packets coming to that address through him, I can't see what could go wrong on such apps if DNS packets are correctly forwarded.

Anyone?

renaudcerrato:
Digging deeper, IPV6 disabled, I'm seeing ICMP packet(s) discarded by Blokada. Could it be related?

```
[IPv4 Header (20 bytes)]
      Version: 4 (IPv4)
      IHL: 5 (20 [bytes])
      TOS: [precedence: 6 (Internetwork Control/ECP)] [tos: 0 (Default)] [mbz: 0]
      Total length: 576 [bytes]
      Identification: 53819
      Flags: (Reserved, Don't Fragment, More Fragment) = (false, false, false)
      Fragment offset: 0 (0 [bytes])
      TTL: 64
      Protocol: 1 (ICMPv4)
      Header checksum: 0x2dbd
      Source address: /203.0.113.1
      Destination address: /203.0.113.2
    [ICMPv4 Common Header (4 bytes)]
      Type: 3 (Destination Unreachable)
      Code: 3 (Port Unreachable)
      Checksum: 0x791b
    [ICMPv4 Destination Unreachable Header (4 bytes)]
      Unused: 0
    [IPv4 Header (20 bytes)]
      Version: 4 (IPv4)
      IHL: 5 (20 [bytes])
      TOS: [precedence: 0 (Routine)] [tos: 0 (Default)] [mbz: 0]
      Total length: 1052 [bytes]
      Identification: 64620
      Flags: (Reserved, Don't Fragment, More Fragment) = (false, true, false)
      Fragment offset: 0 (0 [bytes])
      TTL: 64
      Protocol: 17 (UDP)
      Header checksum: 0xc25f
      Source address: /203.0.113.2
      Destination address: /203.0.113.1
    [UDP Header (8 bytes)]
      Source port: 53 (Domain Name Server)
      Destination port: 44335 (unknown)
      Length: 1032 [bytes]
      Checksum: 0xdc64
    .payload   <--   Thread[tunnel-11,5,main]
```

[...]

PR #1225, opened 2026-08-25
Retention rescue notification + winback tap (#300)

kar:
App half of blokadaorg/issue-tracker#300 (v1 per balboah's flow table). Pairs with api PR blokadaorg/api#55 and translate PR blokadaorg/translate#427. v6 only; Family untouched.

## What

- **`account_rescue` FCM event → `accountRescue` notification** — on receipt the app refreshes its account (10s bound), computes days left from `active_until` (drops if expired or >30 days out — i.e. renewed after send), fetches the all-time blocked count (optional, only if >0), and renders a localized `{title, body}` body at the `schedule_hint` hour: *"Your protection ends soon"* / *"Blokada has blocked N ads and trackers since you set it up. Protection on M devices ends in X days."* (short form without the count). No Adapty re-verification; the 24h FCM TTL plus the account refresh cover staleness.
- **Native prototypes** on iOS and Android (id 27), with schedule-time payload validation on both.
- **Tap → subscription management** — iOS presents the in-app StoreKit `showManageSubscriptions` sheet (`#available(iOS 15)`, URL fallback, every fallback logged); Android opens `play.google.com/store/account/subscriptions?package=org.blokada.sex`. Link templates are now per-platform with a pure, exhaustively tested `selectLinkTemplate`. Tap side-effects that need foreground modules run through a deferred, retried (`20×500ms`, fire-and-forget) action because `NotificationModule` starts before `LinkModule`.
- **Expired-notification tap → `winback` placement** — `Placement.winback`; dismisses an on-screen `accountExpired` modal first, then `openPaymentScreen(placement: winback)`. Previously this tap did nothing.

## Merge gates (in order)

1. ~~translate#427 merged + re-sync~~ — done in 70730cfe (submodule bumped, generated files regenerated from source).
2. **Adapty dashboard**: a placement with id exactly `winback` (both v6 apps) with the Apple/Google win-back offer attached. Without it the expired tap shows the "payment temporarily unavailable" sheet — a regression vs. main, so this gates the *release*, not just the merge.
3. api#55 deployed first (nothing arrives without it).

## On-device sign-off (not run — needs an unlocked device + sandbox account)

iPhone (v6 `Dev`, iOS ≥ 15):
1. Cancel the sandbox subscription so `active_until` is <30 days out; send `account_rescue` to the account via the realtime `/account-expiry` handler (`event_type=account_rescue`, `extras={"devices":"2"}`), once app-killed and once backgrounded → notification with a non-zero count and "Protection on 2 devices ends in N days" (N is one day longer than the store expiry: `active_until` carries the backend's 24h grace — accepted).
2. Same with `schedule_hint` = next hour, then open the app before that hour → still fires at the hour (dismiss-on-foreground only clears delivered notifications).
3. Tap warm and cold → StoreKit sheet in-app; check the device log for `showManageSubscriptions failed` / `No foreground-active scene`.
4. Let the sub expire, kill the app, tap the expired notification → `winback` paywall opens. **Test the cold-start path specifically**: the expired modal is stashed until foreground and can present *after* the paywall; if it covers the paywall that is the known residual to decide on.
5. Foreground receipt at the hinted hour → suppressed, no crash.

Android (`org.blokada.sex`): steps 1, 3 (Play subscriptions page in the Play app, not a browser tab), 4.
Both: `make appium-test` 11/11 (paywall golden mismatch on a non-CI screen size is environmental).

## Known / deferred

- Copy says "ads and trackers"; balboah's table said "ads" — @balboah confirm before #427 merges (a later change costs another translate round). "Protection on 1 devices" plural is also copy-fixed.
- Blocked count is an unformatted integer (`intl` is not a `common` dependency; adding `intl: 0.20.2` enables `NumberFormat`).
- Play link lands on the subscriptions list, not the specific sub (needs the sku, unknown at tap time).
- Pre-existing bug found while extracting the link selector, pinned by a test, not fixed: blokadaorg/issue-tracker#341 (Android knowledge-base link resolves to the iOS one).

## Verification

`make -C common test-local` 435/435; analyzer at the 442-issue baseline (no new issues, errors-only gate); iOS `Dev` arm64 simulator build and `make build-android-six-debug` green.

https://claude.ai/code/session_019jZQTfdiZFeRcZEWApW4nd


common/lib/src/features/link/domain/actor.dart
```diff
@@ -22,8 +24,13 @@ final _linkTemplates = {
...
   LinkTemplate(
       LinkId.privacyCloud, null, null, "https://go.blokada.org/privacy_cloud"),
-  LinkTemplate(LinkId.manageSubscriptions, null, null,
+  // Subscriptions are managed by whichever store sold them, so this one link
+  // has to differ per platform (iOS also gets an in-app StoreKit sheet, see
+  // StageBinding.doOpenLink).
+  LinkTemplate(LinkId.manageSubscriptions, PlatformType.iOS, null,
       "https://apps.apple.com/account/subscriptions"),
+  LinkTemplate(LinkId.manageSubscriptions, PlatformType.android, null,
+      "https://play.google.com/store/account/subscriptions?package=org.blokada.sex"),
```
kar, line 36: Should we not hardcode package id, in case this feature includes family flavor later?

common/lib/src/features/notification/domain/actor.dart
```diff
@@ -271,6 +272,18 @@ class NotificationActor with Logging, Actor {
...
       }
 
+      if (id == NotificationId.accountRescue) {
+        // The whole point of the rescue push is to get the user back to the
+        // store subscription screen, so the tap goes straight there instead of
+        // opening any in-app route.
+        await _queueNotificationAction(
+          m,
+          trigger: "notificationTapped:accountRescue",
+          action: (m) => _stage.openLink(LinkId.manageSubscriptions, m),
```
kar: Check if it's possible to loose this navigation half way in case of cold start for example.

kar:
@copilot review this PR

common/lib/src/features/link/domain/actor.dart
kar: Resolved by flavor-specific templates (Copilot, f48de0f) — the file already selects templates per flavor and there is no runtime package-id accessor in EnvActor. That commit broke the unknown-platform catch-all test though (a flavored android entry won on the flavor clause); fixed the selector precedence in 0d40cbf: a template naming another platform is never a candidate. Same bug was #341 for knowledgeBase, now fixed here too.

common/lib/src/features/notification/domain/actor.dart
kar: Traced it: natively both sides queue commands until the Flutter bridge is ready (CommandsBinding.swift readyTimeout 25s, CommandBinding.kt queue), then Dart parks the action until foreground and retries 20×500ms because LinkModule starts after NotificationModule. Copilot's test (f48de0f) pins the park-then-drain part. Remaining loss window: foreground modules taking >10s → logged and dropped; nothing replays from AppLaunchContext.notificationId. If that should be closed rather than bounded, re-arm the pending action on final failure so the next foreground transition retries — say so and I'll add it.

PR #1235, opened 2026-08-31
feat(family): confirm before an incoming link re-homes a device

balboah:
Closes the gap where a `link_device` link, QR code or intent moved this device onto the account in the incoming token with no user confirmation.

Ref: blokadaorg/issue-tracker#346

## Problem

Two independent defects, both on `main` before this change:

1. **The link committed before it was validated.** `LinkActor.link` called `AuthActor.useToken` (which persists `CurrentToken` and fires `onTokenRefreshed`) *before* `setThisDeviceForLinked` could throw. A rejected link had already overwritten local state, which is why linking with a token from an inactive account showed an error and linked anyway. An ordering defect, not a heartbeat one.
2. **The already-linked guard could not fire across accounts.** It fetched the device list with the *incoming* token and asked that account whether it knew this device's tag, which it never does. The guard only ever fired when re-scanning a different device's QR from the same account.

## Change

- Incoming links park in `PendingLinkValue` (in memory, so an unanswered link does not survive a restart). Nothing is persisted until the user answers.
- The re-home guard compares against the locally persisted `ThisDevice`, and `setThisDeviceForLinked` requires `confirmed: true`, which only the post-confirmation path passes. Any future caller fails closed.
- A confirmation with a red confirm action is shown when the device is already set up. A genuinely fresh device still links with no friction, so child onboarding is unchanged.
- The dialog resolves from `FamilyHomeScreen`, which has a `BuildContext` and cannot be silently deferred the way a `StageModal` is when the stage is not yet foreground. Barrier dismissal and unmount both count as cancel.
- Input parsing accepts the canonical URL with a `token` parameter or the bare token `CommandActivity` delivers, both required to be structurally a JWT. This replaces `tag=`/`name=` parsing that never matched a real link.
- The fixed 3s sleep in `cmdLink` is gone; it started before `runApp` and was never a readiness signal.

### The confirmation predicate

ORs four signals: `WasSetUp`, an active account, `ThisDevice`, `CurrentToken`.

`WasSetUp` is new and persisted, set on a completed link and on parent setup, and never cleared on token expiry. It is load-bearing: a failed token refresh clears both `ThisDevice` and `CurrentToken`, and a linked child holds a `libre` account, so without it a device with plenty to lose reads as fresh. An unresolved account also counts as set up, because `AccountStore.type` reports `libre` before it loads and that is indistinguishable from a real free account.

`FamilyPhase`, `linkedMode` and `DnsPerm` are deliberately not used. They default to the "not set up" value early in boot and would fail open.

## Entry channels

iOS universal link, iOS and Android QR, Android `CommandActivity` intent, and the macOS paste sheet all route through `requestLink`. `CommandActivity` keeps its export and intent filters, so App Store and go.blokada.org links keep working.

## Also fixed

The enrolment token reached the shareable support log in cleartext at info level from two sites that bypassed the `sensitive:` obfuscation. That obfuscation now starts on in release rather than being switched on once `ConfigActor` resolves the log level, which a cold start could beat.

## Verification

463 tests pass, `flutter analyze` reports 0 errors.

On a physical Android device: fresh install links with no dialog, an already-linked child prompts, cancel leaves `this_device` untouched in `shared_prefs`, confirm re-homes, a cold start still prompts, and a barrier dismissal releases the link so a re-scan prompts again.

On a physical iPhone: a reinstalled parent, whose account survives in the keychain while local device state does not, now prompts. That case was found on device and is what the `WasSetUp` and account signals exist for.

## Follow-ups, not in this PR

- `AuthActor._refreshToken` treats any exception, including a network error, as an expired token and discards it (it already carries a `// TODO: may be too aggressive`). That is the root cause of the state-wipe this PR defends against, and changing when auth discards a token deserves its own change.
- A second link parked while the dialog is open stacks a second identical dialog. No privilege gain, but the UX is wrong.
- Inactive-account enforcement at link time is deliberately out of scope.

https://claude.ai/code/session_01EmNq3GBfy9hU3Q8bGk6xSF

### DIFF
--- a/common/lib/src/app_variants/family/module/auth/actor.dart
+++ b/common/lib/src/app_variants/family/module/auth/actor.dart
@@ -28,7 +28,7 @@ class AuthActor with Logging, Actor {
   _recheckToken(Marker m) async {
     final token = await _currentToken.fetch(m);
     if (token != null) {
-      log(Markers.auth).i("current token is $token");
+      log(Markers.auth).i("found a current token");
       try {
         useToken(token, m);
         startHeartbeat();
--- a/common/lib/src/app_variants/family/module/device_v3/actor.dart
+++ b/common/lib/src/app_variants/family/module/device_v3/actor.dart
@@ -30,6 +30,7 @@ class AlreadyLinkedException implements Exception {
 class DeviceActor with Logging, Actor {
   late final _devices = Core.get<DeviceApi>();
   late final _thisDevice = Core.get<ThisDevice>();
+  late final _wasSetUp = Core.get<WasSetUp>();
   late final _nextAlias = Core.get<NameGenerator>();
   late final _profiles = Core.get<ProfileActor>();
 
@@ -79,7 +80,12 @@ class DeviceActor with Logging, Actor {
           m);
 
       await _thisDevice.change(m, newDevice);
+      await _wasSetUp.change(m, true);
       devices = await _devices.fetch(m);
+    } else {
+      // Backfill for installs predating WasSetUp: a device that is already in
+      // the account is set up, whatever the flag currently says.
+      await _wasSetUp.change(m, true);
     }
   }
 
@@ -130,21 +136,19 @@ class DeviceActor with Logging, Actor {
     return d;
   }
 
-  setThisDeviceForLinked(DeviceTag tag, String token, Marker m) async {
+  setThisDeviceForLinked(DeviceTag tag, String token, Marker m,
+      {bool confirmed = false}) async {
     final devices = await _devices.fetchByToken(tag, token, m);
     final d = devices.firstWhereOrNull((it) => it.deviceTag == tag);
     if (d == null) throw Exception("Device $tag not found");
 
-    final thisDevice = await _thisDevice.now();
+    final thisDevice = await _thisDevice.fetch(m);
     if (thisDevice?.deviceTag == tag) return;
 
-    // If we were linked to another device, we cant change it now
-    if (thisDevice != null) {
-      // If the linked device got deleted, allow to re-link
-      final confirmedThisDevice = devices
-          .firstWhereOrNull((it) => it.deviceTag == thisDevice.deviceTag);
-      if (confirmedThisDevice != null) throw AlreadyLinkedException();
-    }
+    // Re-homing onto another account needs the user's answer. Compared against
+    // locally persisted state, not the incoming token's device list.
+    if (thisDevice != null && !confirmed) throw AlreadyLinkedException();
+
     await _thisDevice.change(m, d);
   }
 
--- a/common/lib/src/app_variants/family/module/device_v3/device.dart
+++ b/common/lib/src/app_variants/family/module/device_v3/device.dart
@@ -20,6 +20,7 @@ class DeviceModule with Module {
   @override
   onCreateModule() async {
     await register(CurrentToken());
+    await register(WasSetUp());
     await register(SelectedDeviceTag());
     await register(SlidableOnboarding());
     await register(NameGenerator());
--- a/common/lib/src/app_variants/family/module/device_v3/value.dart
+++ b/common/lib/src/app_variants/family/module/device_v3/value.dart
@@ -10,6 +10,13 @@ class SelectedDeviceTag extends NullableAsyncValue<DeviceTag> {
   }
 }
 
+/// True once this device has been set up, as a linked child or as a parent
+/// with its own controller device. Never cleared on token expiry, so a wiped
+/// token cannot make a set up device look fresh to the link confirmation.
+class WasSetUp extends BoolPersistedValue {
+  WasSetUp() : super("was_set_up");
+}
+
 class SlidableOnboarding extends BoolPersistedValue {
   SlidableOnboarding() : super("slidable_onboarding");
 }
--- a/common/lib/src/app_variants/family/module/family/command.dart
+++ b/common/lib/src/app_variants/family/module/family/command.dart
@@ -13,10 +13,8 @@ class FamilyCommand with Command {
   Future<void> cmdLink(Marker m, dynamic args) async {
     final url = args[0] as String;
 
-    // When entering from a camera app qr code scan, this will be called
-    // by the OS very early, and since onStartApp is executed async to not
-    // block the UI thread, we need to wait for the app to be ready.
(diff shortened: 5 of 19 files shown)


balboah:
@codex please re-check

Addressed the P1 on `was_set_up` not being backfilled for existing installations. The flag is now also set:
- in the token-expiry handler, before it clears `ThisDevice` (holding a token at all means the device was set up, and this is the path that wipes the other signals)
- in `DeviceActor.reload` when the device is already present in the account, which covers upgraded parents

Regression test added: `token expiry backfills the setup flag on an upgraded install` (verified it fails without the fix). 464 tests pass, analyze clean.

balboah:
@codex please re-check

Both findings verified against the code and fixed:

**P1 (bind confirmation to the displayed token).** Confirmed: `confirmPendingLink` read `pendingLink.present` at tap time, so a newer link could be committed by approving the older prompt. `confirmPendingLink` and `cancelPendingLink` now take the token the prompt was raised for and no-op if the pending value has moved on. The captured token is threaded through the dialog callbacks.

**P2 (backfill only when a token existed).** Confirmed: `useToken` calls `_startMonitoringTokenExpiry` before `_currentToken.change`, so a rejected or unreachable candidate token reaches the expiry callback with nothing persisted. The backfill is now gated on `CurrentToken` or `ThisDevice` actually being set, which keeps a fresh install's first link friction-free. The callback is also `async` now so the flag write is ordered before the wipe rather than racing it.

Regression tests added, both verified to fail without their fix:
- `confirming a stale prompt does not commit a newer link`
- `a rejected first link does not mark a fresh device as set up`

466 tests pass, analyze clean.

balboah:
@codex please re-check

Verified and fixed. `_promptedFor` is assigned synchronously before the `needsConfirmation` await, so a second resolve carrying a different token passed the guard and both continued to a dialog. `_resolvePendingLink` now re-reads the pending value after the predicate and returns unless it still owns the token, so two links raise one prompt for whichever is current.

No new test: `FamilyHomeScreen` has no widget-test harness today, and standing one up for a two-line ordering guard is disproportionate. Flagging that as a known coverage gap rather than pretending it is covered.

466 tests pass, analyze clean.

balboah:
@codex please re-check

Confirmed, and this is the third finding on the same mechanism (token binding, then post-await ownership, now stale-dialog dismissal), so I fixed the source rather than this instance.

`requestLink` now drops a second link while one still awaits an answer, instead of replacing the pending value under an open prompt. That makes "at most one pending link, at most one dialog" true by construction, so there is no stale dialog to dismiss and no need for machinery to reach into an open route. If the user meant the newer link they re-scan; the parked one is released on cancel, dismissal or unmount.

The token binding and the post-await ownership check stay as defence in depth.

Test `a second link while one awaits an answer is dropped` added; the stale-prompt test now drives the pending value directly since `requestLink` no longer produces that state. 467 tests pass, analyze clean, and all 5 CI checks were green on the previous head.

balboah:
@codex please re-check

Confirmed against `core/value.dart` and fixed. `NullableAsyncValue.change` assigns `_value`, then awaits `save?.call(...)` (which yields even when `save` is null), and only sets `_resolved = true` afterwards, while `present` returns null until `_resolved`. So two links arriving before the value had ever resolved both passed the null check.

`requestLink` now claims into an actor-owned `String? _claimedLink`, checked and assigned synchronously so nothing can interleave between them, and the answer paths key off that claim rather than `present`. A failed commit releases the claim, so a rejected link cannot wedge the slot.

Test `concurrent links before the value resolves still claim once` added. It reproduces the race: it fails against the previous `present`-based check and passes with the claim.

468 tests pass, analyze clean.

balboah:
@codex please re-check

Confirmed and fixed. `_claimedLink` was released before `_commitLink`, and that commit persists in two steps (`useToken` writes `CurrentToken`, then `setThisDeviceForLinked` writes `ThisDevice`), so a link accepted in between could pair one family's token with another's device tag. The claim is now held across the commit and released in a `finally`, so a rejected link still cannot wedge the slot.

Worth noting the reachable shape: on an already set up device both links prompt, and nobody answers a second dialog inside the commit window, so that path is unreachable. The real exposure was a fresh device, where neither link needs confirmation and both auto-commit.

Test `a link arriving mid-commit is dropped` added. It is deterministic rather than timing-based: `useToken` is held open with a Completer while the second link is requested. Verified it fails without the fix.

469 tests pass, analyze clean.

balboah:
@codex please re-check

Device testing on a physical Android found a regression my unit tests were hiding, now fixed.

**What broke:** on a genuinely fresh install, a cold start via link showed the re-home confirmation. `needsConfirmation` treated `_account.account == null` as "set up", but after `_startCoreBootstrap` a null account legitimately means "no account on this device", not "not loaded yet". The unit test passed only because it stubbed `account` with a non-null mock, encoding an assumption that is false on the cold-start path.

**Fix:** ask persistence directly instead of guessing. New `AccountStore.hasActivePersistedAccount`, reading the same `keyAccount`/`isBackup` store that `_startCoreBootstrap` reads, so the answer no longer depends on whether startup has resolved the account. That keeps the iOS case (account survives a reinstall in the keychain while device state is wiped) while letting a fresh install link silently.

The bad test is replaced by `a stored account needs confirmation before startup loads it`, which drives the persisted signal rather than the loaded one.

Verified on device, all six reproducible cases: fresh install links with no dialog; linked child prompts and cancel leaves `this_device` untouched; confirm re-homes; cold start prompts; barrier dismissal releases the claim; re-firing the same link prompts again.

Note on the one-at-a-time rule: two rapid intents are not reproducible through Android's `CommandActivity`, which is `singleTask` and reads `intent.data` only in `onCreate`, so only one command reaches Flutter. That path stays covered by unit tests rather than hardware.

469 tests pass, analyze clean.

Still outstanding and unaddressed: your macOS paste-sheet finding about `Navigator.pop()` racing the pending-link listener.

balboah:
@codex please re-check

balboah:
@codex please re-check

kar (approved):
I can't quite tell if all major codex issues are addressed in latest version, some of its comments still show in diffs. Code wise, my concern would be timing/race around the linking logic, as I had some problems with it in the past, related to how AsyncValue works. But I see there is some (new) attempt to address those here, without sleep also which is always good.

PR #1240, opened 2026-09-02
Revert cached_network_image 4.0.0 (breaks on-device navigation)

kar:
The iOS physical-device smoke went red on **every** spec on the commit that merged #1238 (`build(deps): bump cached_network_image 3.4.1 → 4.0.0`), green on the commit immediately before it. Failure is identical across all 11 specs: the Settings screen never renders after the home gear is tapped (`element ("~automation.screen_settings") still not existing after 15000ms`), so no spec can seed its account — a build/app break, not a flaky element. [Failing run](https://github.com/blokadaorg/blokada/actions/runs/33667221904).

`cached_network_image` 4.0.0 is a brand-new major (published 2026-08-28; Baseflow's GitHub shows no 4.0.0 release), pulls in the new `material_ui` / `cupertino_ui` split-out packages, and bumps `cached_network_image_platform_interface` to 5.0.0. CI is build + lint only, so it merged green; the device smoke is the first thing that exercised it.

## What this does
- Reverts #1238 (pubspec + lock back to `cached_network_image` 3.4.1; `material_ui` / `cupertino_ui` gone from the lock).
- Tightens the constraint to `">=3.3.1 <4.0.0"` so `pub get` can't re-pick 4.x.
- Holds the 4.x major in Dependabot: an `ignore` on `cached_network_image` semver-major plus a `flutter-dependencies` group exclusion (same trap as pigeon — a group recreation would otherwise re-widen it). Drop both once 4.x is dep-validated on device.

Not related to the retention feature (issue-tracker#300) — that code is intact on `main`; this just unblocks cutting a release build with a green smoke.

https://claude.ai/code/session_019jZQTfdiZFeRcZEWApW4nd

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,6 +23,7 @@ updates:
         # with the ignore when the pigeon 26 native migration is scheduled.
         exclude-patterns:
           - "pigeon"
+          - "cached_network_image"
         update-types:
           - "minor"
           - "patch"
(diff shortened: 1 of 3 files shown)


kar:
false alarm, turned out it was another OS popup on device

Issue #69, opened 2017-11-23
HTTPS on website


edent:
Love the app - but I'm a bit concerned about its security.  The configuration is only downloaded over HTTP.  Ideally, the `http://blokada.org/api/` URLs should be protected with `https`.

Additionally, in http://blokada.org/api/20/dev/filters.txt the following should be replaced by their HTTPS equivalents.

http://pgl.yoyo.org/adservers/serverlist.php?hostformat=hosts&showintro=0 -> https://pgl.yoyo.org/adservers/serverlist.php?hostformat=hosts&showintro=0

kar:
Definitely

hoshsadiq:
CloudFlare can be used to enable https: https://hackernoon.com/set-up-ssl-on-github-pages-with-custom-domains-for-free-a576bdf51bc

hoshsadiq:
@kar one more thing, would be nice to enable http to https redirect.

kar:
Yes absolutely. Once I update all the clients and there is no unexpected problems, I'll do so.

PR #1243, opened 2026-09-03
fix(account): show the account-expired notification once per lapse

kar:
## Why

blokadaorg/issue-tracker#344 — an expired v6 account gets the "account expired" notification
more than once, days apart.

Every `account_expiry` FCM event runs `_maybeShowImmediateAccountExpiryNotification`, whose
only guards were family-linked mode and "the expiry is in the future". A repeat event for an
already-expired account showed the notification again.

The server-side cause is fixed in blokadaorg/api#69. This is the client half: it holds
regardless of how often the server dispatches for one lapse.

## What

**One notification per expiry, keyed by the expiry itself.** `JsonAccRefreshMeta` gains
`expiryNotifiedFor` — the expiry the notification announced, not the time it was shown. The
immediate path skips when it matches the expiry now in hand. Instants are compared, not the
raw strings, so a formatting change on the server cannot defeat the match.

The next lapse carries a different `active_until` by construction, so nothing needs re-arming
on renewal and there is no window to tune. This depends on `active_until` staying stable for a
lapse, which blokadaorg/api#69 fixed and is merged — a future change that restamps a lapsed
account's expiry would silently re-open this bug.

The lapse is claimed before the `show()` await, since nothing serializes the FCM handler and
a burst would otherwise slip several notifications past the check. A failed `show()` gives the
claim back, so a mark can never outlive a notification that was not delivered.

**Behaviour change:** an account with no parsable `active_until` no longer shows the
notification at all. It previously fell through the same branch as an expired one and
announced an expiry with no date.

**Also fixed:** metadata now loads through `_ensureMetadataLoaded`, called from
`onAccountExpiryEvent` as well as `init()`. A background FCM can land before `init()` ran —
`handleFcmEvent` only requires `_account.account != null`, which the bootstrap preload can
satisfy first — and the guard must not read defaults or save over the stored
`previousAccountType` / `seenExpiredDialog`. A corrupt blob falls back to fresh metadata
instead of throwing before the account is even refreshed. `_updateTimer` now awaits its
`show()`, so a scheduling failure is logged rather than dropped on the floor.

## Not in scope

An earlier revision also suppressed the immediate notification when an OS notification had
been scheduled for the same expiry. That is removed: after keying on the expiry it no longer
catches the pair it was written for, and it was the only part of the guard that could leave a
real lapse unannounced. Reasoning in `84918106`.

The duplicate it was aiming at is a stale alarm, not a dedupe problem, and fixing it properly
needs Android-side work — an alarm-only cancel, a delivery marker in
`NotificationAlarmReceiver.onReceive`, and re-arming in `BootReceiver`, which currently only
logs. That is one coherent change in one layer and belongs in its own issue.

## Tests

Six cases in `refresh_test.dart` plus three `JsonAccRefreshMeta` cases, including stored
metadata that predates the field. The load-bearing one is `notifies once for repeated events`.

Verified by falsification: neutralising the expiry match fails the two dedupe tests and the
concurrency one, comparing raw strings instead of instants fails only the offset test,
dropping the claim restore fails only the retry test, and dropping the schedule catch fails
only the arming test.

`make test` green (479 tests). `flutter analyze` reports the same 441 issues as `main`, zero
errors.

## Not verified here

Push delivery itself is not covered by the Appium suite. Worth a device check before release:
let an account lapse, send two `account_expiry` pushes a minute apart, confirm one
notification.

https://claude.ai/code/session_01NC1HnqYRypEd6ZGhAwgSyM

### DIFF
--- a/common/lib/src/core/config.dart
+++ b/common/lib/src/core/config.dart
@@ -8,9 +8,10 @@ class CoreConfig {
   // How long one lapse owns the "account expired" notification. The backend
   // restamps active_until on every repeat webhook for a lapsed account, so the
   // expiry cannot key the guard; this floor does. Longer than any restamp
-  // burst, far shorter than the shortest paid plan, and a renewal clears the
-  // mark anyway.
-  Duration accountExpiredNotificationCooldown = const Duration(days: 7);
+  // burst, and short enough that it cannot reach the next lapse of the shortest
+  // billing period sold, which is weekly. A renewal clears the mark anyway, but
+  // only if the app sees it, so the floor does not lean on that.
+  Duration accountExpiredNotificationCooldown = const Duration(hours: 72);
   Duration deviceRefreshCooldown = const Duration(seconds: 60);
   Duration plusLeaseRefreshCooldown = const Duration(seconds: 60);
   Duration plusGatewayRefreshCooldown = const Duration(seconds: 60);
--- a/common/lib/src/core/config.dart
+++ b/common/lib/src/core/config.dart
@@ -4,6 +4,13 @@ class CoreConfig {
   Duration appStartFailWait = const Duration(seconds: 5);
   Duration accountExpiringTimeSpan = const Duration(seconds: 30);
   Duration accountRefreshCooldown = const Duration(seconds: 60);
+
+  // How long one lapse owns the "account expired" notification. The backend
+  // restamps active_until on every repeat webhook for a lapsed account, so the
+  // expiry cannot key the guard; this floor does. Longer than any restamp
+  // burst, far shorter than the shortest paid plan, and a renewal clears the
+  // mark anyway.
+  Duration accountExpiredNotificationCooldown = const Duration(days: 7);
   Duration deviceRefreshCooldown = const Duration(seconds: 60);
   Duration plusLeaseRefreshCooldown = const Duration(seconds: 60);
   Duration plusGatewayRefreshCooldown = const Duration(seconds: 60);
--- a/common/lib/src/core/config.dart
+++ b/common/lib/src/core/config.dart
@@ -5,13 +5,13 @@ class CoreConfig {
   Duration accountExpiringTimeSpan = const Duration(seconds: 30);
   Duration accountRefreshCooldown = const Duration(seconds: 60);
 
-  // How long one lapse owns the "account expired" notification. The backend
-  // restamps active_until on every repeat webhook for a lapsed account, so the
-  // expiry cannot key the guard; this floor does. Longer than any restamp
-  // burst, and short enough that it cannot reach the next lapse of the shortest
-  // billing period sold, which is weekly. A renewal clears the mark anyway, but
-  // only if the app sees it, so the floor does not lean on that.
-  Duration accountExpiredNotificationCooldown = const Duration(hours: 72);
+  // How long after an expiry the OS notification scheduled for it is assumed to
+  // have been delivered, so the push announcing the same expiry stays quiet.
+  // Bounded because arming an alarm is not delivering one: Android drops
+  // pending alarms on reboot and nothing re-arms them for an account that has
+  // already expired, leaving the push as the only thing that can announce that
+  // lapse. Past this window it does, rather than trusting the record.
+  Duration accountExpiryScheduledGrace = const Duration(hours: 6);
   Duration deviceRefreshCooldown = const Duration(seconds: 60);
   Duration plusLeaseRefreshCooldown = const Duration(seconds: 60);
   Duration plusGatewayRefreshCooldown = const Duration(seconds: 60);
--- a/common/lib/src/core/config.dart
+++ b/common/lib/src/core/config.dart
@@ -4,14 +4,6 @@ class CoreConfig {
   Duration appStartFailWait = const Duration(seconds: 5);
   Duration accountExpiringTimeSpan = const Duration(seconds: 30);
   Duration accountRefreshCooldown = const Duration(seconds: 60);
-
-  // How long after an expiry the OS notification scheduled for it is assumed to
-  // have been delivered, so the push announcing the same expiry stays quiet.
-  // Bounded because arming an alarm is not delivering one: Android drops
-  // pending alarms on reboot and nothing re-arms them for an account that has
-  // already expired, leaving the push as the only thing that can announce that
-  // lapse. Past this window it does, rather than trusting the record.
-  Duration accountExpiryScheduledGrace = const Duration(hours: 6);
   Duration deviceRefreshCooldown = const Duration(seconds: 60);
   Duration plusLeaseRefreshCooldown = const Duration(seconds: 60);
   Duration plusGatewayRefreshCooldown = const Duration(seconds: 60);
--- a/common/lib/src/platform/account/refresh/json.dart
+++ b/common/lib/src/platform/account/refresh/json.dart
@@ -4,20 +4,43 @@ class JsonAccRefreshMeta {
   late AccountType? previousAccountType;
   late bool seenExpiredDialog;
 
+  /// When the immediate expiry notification was last shown, ISO-8601.
+  ///
+  /// Not keyed by the expiry itself: the backend restamps active_until to now
+  /// on every repeat webhook for a lapsed account, so a per-expiry key would be
+  /// a different value every time. A time floor is what actually dedupes.
+  late String? expiryNotifiedAt;
+
+  /// The expiry the OS notification was last scheduled for, ISO-8601.
+  ///
+  /// The scheduled notification and the immediate one are the same lapse seen
+  /// twice, so the immediate path checks this too.
+  late String? expiryScheduledFor;
+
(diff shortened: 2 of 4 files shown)


kar:
Second commit (`6f10f5b5`) applies review findings. Each fix was verified by reverting it in isolation and confirming exactly the intended test fails.

- **Concurrency (the important one).** Nothing serializes the FCM handler, and `onAccountExpiryEvent` awaits a metadata load, an account fetch and a sync before reaching the guard — so two pushes arriving together both read a null `expiryNotifiedAt` and both notified. The lapse is now claimed synchronously, before the first await.
- **`_updateTimer` recorded a schedule that may never have landed.** `show()` was not awaited, so a failing platform channel would still have set `expiryScheduledFor` and suppressed the immediate notification — the fallback for exactly that case — for a week. Now awaited, and a failure records nothing.
- **Clearing the mark on `expiring`, not only `active`.** An `active_until` restamped a few seconds into the future lands in `expiring` on a slow device clock, which re-armed the guard mid-lapse and produced a second notification.
- **Unguarded `jsonDecode`.** Since the load moved to the top of `onAccountExpiryEvent`, a corrupt blob would have thrown before the account was even refreshed. It now falls back to fresh metadata, shares one in-flight load, and re-checks the loaded flag afterwards so it cannot resurrect a stale blob over a newly created account.
- Timestamps are written in UTC (`expiryNotifiedAt` was local with no offset while `expiryScheduledFor` was UTC, so the two were not comparable).
- `resolveAccountExpirySchedule` had no callers left and is removed with its tests.

482 tests green, analyzer unchanged from `main` (441 issues, zero errors), CI 5/5.

One review point deliberately not changed: a missing or unparseable `active_until` no longer notifies. `ActivateAccount` always writes `active_until`, and a zero time still serializes as a parseable date, so there is no path where a real lapse arrives without one.

balboah (changes requested):
Three things, the first of which should simplify the change considerably.

**1. Key on the expiry, not on elapsed time.**

The floor is only needed because `expiryNotifiedAt` records *when* we notified rather than *what* we notified about. That is what makes the mark need clearing in `_updateTimer`, makes the clear depend on the app seeing the renewal, and then bounds the floor below the weekly plan at 72h.

Storing the expiry we notified for collapses all of it: once per expiry, no window, no clearing, and `accountExpiredNotificationCooldown` goes away. The next lapse carries a different `active_until` by construction, so nothing needs re-arming. It also makes the client indifferent to how often the server dispatches for one expiry, which is the property worth having, since the remaining duplicate sources (FCM redelivery, Adapty retrying a 500, the scheduled-plus-immediate pair) all carry the same `active_until`. The only server behaviour it depends on is that value staying stable for a lapse, which blokadaorg/api#69 fixed.

Compare parsed instants rather than raw strings, so a server-side formatting change cannot defeat the match.

**2. `alreadyScheduled` treats an armed alarm as a delivered notification.**

Independent of the keying. The app records that it armed an alarm, then treats that record as proof the user saw something. Android drops alarms on reboot (`alarmManager.set(RTC_WAKEUP, ...)`, `NotificationService.kt:110`), `BootReceiver.kt` only logs, and `_updateTimer` will not re-arm an already-expired account because `getNextDate()` returns null. So the record can be true while nothing was shown.

Today the FCM push covers that gap, since the current code shows the notification immediately for any past expiry. This PR makes the record suppress the push, which removes the only compensation for a lost alarm. The scope is one lapse, a later expiry after a renewal still notifies, but that lapse goes unannounced. iOS is unaffected, pending `UNUserNotificationCenter` requests survive reboot.

Nothing rescheduling after a reboot is pre-existing and its own problem. The narrow fix here is to not let arming imply delivery: bound this branch to a window covering the gap between the OS firing and the FCM landing.

**3. A failed `show()` still burns the mark.**

`expiryNotifiedAt` is set before `await _notification.show(...)`. If that throws (iOS `doShow` fails via `completion(.failure)`), `_saveMetadata` is skipped but the in-memory mark survives and is persisted by the next `_saveMetadata`, either the `syncAccount` tail or `_updateTimer`. `_updateTimer` already avoids recording a schedule that failed, and its comment names the immediate notification as the fallback for that case, so the two paths should agree. Restoring the previous value in a catch keeps the claim-before-await property.

`_ensureMetadataLoaded` is a real fix independent of the guard, worth keeping however the keying lands.

kar:
All three applied in `d1164127`. Each verified by reverting it in isolation and confirming exactly the intended tests fail.

**1. Keyed on the expiry.** `expiryNotifiedAt` → `expiryNotifiedFor`, matched against the expiry in hand as instants rather than strings. `accountExpiredNotificationCooldown` is gone, and so is the clearing in `_updateTimer` — the next lapse re-arms itself. Net simplification, and the client no longer cares how often the server dispatches for one lapse.

Worth stating plainly, since it is the trade: the client now depends on `active_until` being stable for a lapse. That was not true before api#69 — the restamping is what forced the floor in the first place — so the guard is only correct against a backend carrying that fix. #69 is merged, and this ships in an app release well after it deploys, so the ordering holds. It does mean a future change that restamps a lapsed account's expiry silently re-opens this bug, which the time floor would have absorbed.

**2. `alreadyScheduled` bounded.** Confirmed the mechanism: `alarmManager.set(RTC_WAKEUP, ...)` at `NotificationService.kt:110`, `BootReceiver` logs and nothing else, and `getNextDate()` returns null for an expired account so `_updateTimer` never re-arms. New `Core.config.accountExpiryScheduledGrace`, 6h.

On the value: the gap this covers is genuinely short. A natural lapse writes nothing server-side, so the two only coincide when a webhook lands around the expiry, which puts the push within minutes of the alarm. Six hours is generous cover for a delayed delivery while keeping the silent window small — consistent with your point on api#70 that a missed announcement is worse than a duplicate. Note it does not fix the case where the alarm was lost *and* the push arrives promptly; both are inside the window, so that lapse still goes unannounced. Fixing that needs delivery to be observable, or a reboot-time reschedule, which is the pre-existing problem you set aside.

**3. Failed `show()`.** Previous value restored in a catch and rethrown, so the claim-before-await property holds and the two paths agree.

Test churn from the keying: `does not renotify when active_until is restamped` and `does not renotify when a restamp lands just in the future` both asserted behaviour that only made sense under the floor — a different `active_until` is now a different lapse by definition. Removed, and replaced by `matches a stored expiry written in another offset` (instants, not strings), `notifies when the scheduled alarm is too old to vouch for`, and `retries after the notification failed to show`. `notifies once for repeated events` is now the regression test.

482 tests green, analyzer unchanged from `main` (441 issues, zero errors).

balboah:
feels a bit silly having our claudes ask each other, since I'm not confident in the remaining point 2 I'll let your claude talk to you and another agent directly.

so approved from that point of view, although I can't mark it approved as there's no new review request

kar:
Point 2 is withdrawn rather than tuned: `84918106` deletes the `alreadyScheduled` branch, `expiryScheduledFor` and `accountExpiryScheduledGrace`. The guard is now one rule.

This corrects something I claimed in my last comment. I said the grace window covered the scheduled-plus-immediate pair. Under the keying from point 1 it mostly does not, and I did not re-derive it when the keying changed. Post api#69 the pair arises two ways:

- **Lapse webhook lands after the expiry** — the normal case. `!a.ActiveUntil.After(now) && !until.After(now)` holds, `ActivateAccount` returns without writing (`account.go:172`), so there is no Firestore change, no dispatch, no push. The alarm fires alone. Nothing to dedupe.
- **Revocation or early lapse, webhook lands while still active** — writes `active_until = now`, a *different instant* than the `T` the alarm was armed for. `isAtSameMomentAs` fails, the branch does not fire, and the user gets the push now plus the stale alarm at `T` later. This is the case the branch existed for, and keying on the expiry took it away.

So what the branch still catches needs a write that leaves `active_until` untouched after it has already passed. Against that it was the only part of the guard that could silence a real lapse, for the reason you gave — which is the wrong side of the trade you drew on api#70.

Your point stands in full, in other words; the window was just the wrong repair for it. A window bounds the risk but cannot remove it, because the client records its own intent and never observes delivery. It also does not help the case that motivated it: alarm lost at reboot, push arriving promptly, both inside any window.

The second bullet is a stale alarm, not a duplicate. Cancelling it needs an alarm-only cancel — `NotificationService.cancel()` also dismisses an already-shown notification, which is presumably why `_updateTimer` only cancels for an active account. That, a delivery marker in `NotificationAlarmReceiver.onReceive`, and re-arming in `BootReceiver` (the `RECEIVE_BOOT_COMPLETED` permission and the `BOOT_COMPLETED` registration are already there, the body is one `Log.i`) are one piece of Android work. Happy to open the issue — say the word and I will, or leave it to you if you would rather scope it.

`_updateTimer` keeps the awaited `show()`, since a scheduling failure used to be dropped silently.

479 tests green, analyzer unchanged from `main`. Falsified again after the removal: neutralising the expiry match fails the two dedupe tests and the concurrency one, raw-string comparison fails only the offset test, dropping the claim restore fails only the retry test.

PR #1250, opened 2026-09-07
fix(android): support chat unusable behind the keyboard on old Android

kar:
Fixes blokadaorg/issue-tracker#152.

## The bug

On Android 9/10 the support chat composer stayed **behind** the soft
keyboard: you could not see what you typed, and could not reach the send
button without dismissing the keyboard first. Enter did not send either.
Newer Android was fine, which is the clue that pinned it down.

## Root cause

`MainActivity` set `FLAG_LAYOUT_NO_LIMITS` on top of
`WindowCompat.setDecorFitsSystemWindows(window, false)`. That flag zeroes the
window's content insets. The Flutter engine shipped with this build
(3.44.6) splits on API level in `FlutterView.onApplyWindowInsets`:

- API >= 30 reads `insets.getInsets(WindowInsets.Type.ime()).bottom` — a typed
  inset that survives the flag, so modern Android worked.
- API < 30 derives the keyboard height from `getSystemWindowInsetBottom()` —
  which the flag zeroes.

So on Android 9 `MediaQuery.viewInsets.bottom` stayed 0, `Scaffold` never
shrank, and flutter_chat_ui's bottom-anchored composer stayed under the
keyboard. No Dart-side change could compensate: Flutter had no other source
of keyboard height.

The manifest also never declared `windowSoftInputMode`, which the Flutter app
template ships as `adjustResize`.

Separately, the composer asked for a multiline keyboard type, which makes
Android IMEs draw a newline key and ignore the send action entirely.

## The fix

1. Drop `FLAG_LAYOUT_NO_LIMITS` and declare `adjustResize`. Edge-to-edge
   still comes from `setDecorFitsSystemWindows`.
2. Re-request insets once the first Flutter frame is out. The first dispatch
   reaches `FlutterView` before the hosting fragment is measured, and
   Flutter's pre-11 heuristic then misreads the navigation bar as a keyboard;
   without this the phantom inset sticks and leaves a band at the bottom of
   every screen.
3. Zero the hardcoded 44dp bottom reserve for short Android phones while the
   keyboard is up. It stood in for the navigation bar Flutter could not see,
   and would now sit as a dead band between composer and keyboard.
4. Give the chat a composer with a plain-text keyboard type and a send
   action, so Enter sends. Focus is handed back after a send, which Flutter
   supports from the submit callback, so the keyboard stays up.

## Verification

Reproduced and fixed on an **Android 9 (API 28) emulator**, the reporter's OS
version. Before: keyboard covers the composer, raw `viewInsets.bottom` stays
0 with the IME open. After: composer sits directly above the keyboard, the
keyboard shows a send key, Enter sends and the keyboard stays open.

| Check | Result |
| --- | --- |
| Android 9, six flavor, chat + keyboard + Enter | fixed |
| Android 9, six flavor, home / settings / account-restore dialog | no band, dialog lifts above the keyboard |
| Android 9, family flavor, launch + layout | unchanged |
| Android 16 (API 36), chat + keyboard + Enter | unchanged, still correct |
| `make test` | 480 passing |
| `flutter analyze` | no new errors or warnings |

New widget test covers the composer's send action and its keyboard type.

## Note for review

On Android 9/10 the navigation bar is now opaque (`navigationBarColor` from
the theme) instead of translucent over app content. That follows from
dropping the flag, and the icons are more legible than before, but it is a
visible change on old phones.

https://claude.ai/code/session_018ms5SZ7Q2UnRtzioUfxPh4

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -50,7 +50,6 @@
             android:label="@string/app_name"
             android:launchMode="singleTask"
             android:screenOrientation="portrait"
-            android:windowSoftInputMode="adjustResize"
             android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -50,6 +50,7 @@
             android:label="@string/app_name"
             android:launchMode="singleTask"
             android:screenOrientation="portrait"
+            android:windowSoftInputMode="adjustResize"
             android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
--- a/android/app/src/main/java/ui/MainActivity.kt
+++ b/android/app/src/main/java/ui/MainActivity.kt
@@ -24,6 +24,7 @@ import androidx.activity.OnBackPressedCallback
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.view.ViewCompat
 import androidx.core.view.WindowCompat
+import androidx.core.view.doOnLayout
 import androidx.lifecycle.lifecycleScope
 import binding.CommandBinding
 import binding.CommonBinding
@@ -100,7 +101,7 @@ class MainActivity : AppCompatActivity() {
     }
 
     /**
-     * Re-dispatches window insets once the Flutter view tree has a real height.
+     * Re-dispatches window insets once the view tree has been measured.
      *
      * Below Android 11 Flutter has no typed ime() inset to read, so it decides
      * whether the window's bottom inset is a keyboard by comparing it against
@@ -109,13 +110,14 @@ class MainActivity : AppCompatActivity() {
      * navigation bar as a keyboard: every screen then keeps a phantom bottom
      * inset, and its content stops short of the navigation bar, until some
      * later inset change happens to correct it. See issue-tracker#152.
+     *
+     * Keyed on this activity's own layout rather than on Flutter's first frame,
+     * which is a process-wide latch: after an activity recreate that latch is
+     * already set, so it would fire back synchronously here, before layout, and
+     * never again.
      */
     private fun requestApplyInsetsOnceLaidOut() {
-        StartupContextService.addFirstFlutterFrameListener {
-            runOnUiThread {
-                window.decorView.post { ViewCompat.requestApplyInsets(window.decorView) }
-            }
-        }
+        window.decorView.doOnLayout { ViewCompat.requestApplyInsets(it) }
     }
 
     private var splashFallbackRunnable: Runnable? = null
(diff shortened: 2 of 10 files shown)


kar:
Review pass follow-up:

- **Fixed in 4ad40f56.** The inset re-dispatch keyed off Flutter's first-frame signal, which is a process-wide latch: after an activity recreate it fired synchronously inside `onCreate` before the decor view was attached, and never again, so the phantom bottom inset returned. It now keys on the activity's own layout, which also drops the strong reference it held in the static listener list. Verified on Android 9 with a dark-mode recreate.
- **Navigation bar colour: tried and reverted.** Setting `navigationBarColor` transparent makes no visible difference on Android 9 with button navigation. I confirmed the transparent value reached the APK (`aapt2 dump`) and the bar area stayed dark, so the strip is not the framework's colour view. Left the theme untouched rather than adding an unverifiable change; the appearance note in the PR description stands.
- **`isSmallAndroid` flipping while typing: not reproducible.** `MediaQuery.size` stayed 411.4x731.4 with the keyboard open and closed on Android 9, and `dumpsys window` shows the window frame unchanged at 1080x1920 with the IME up — `adjustResize` does not shrink the frame here, the keyboard arrives as an inset. So the height heuristic does not flip mid-typing.

kar:
Two device-testing findings, both fixed:

**Stuttery chat while the keyboard animates** — caused by this PR. Routing the composer through `builders` replaced flutter_chat_ui's `const Composer()` with a fresh instance per rebuild. The library re-measures itself from `didUpdateWidget` and pushes the height into `ComposerHeightNotifier`, which the message list consumes as bottom spacing. The chat screen already rebuilt once per keyboard-animation frame (it reads MediaQuery for its top padding), so each of those frames now re-measured and re-laid-out the list. Fixed in 79146d8a by building the composer once and handing back the identical widget.

**Opening greeting arriving twice** — pre-existing, not from this PR, but fixed here since it is in the same screen. `startSession` clears the session id before the create call and stores the new one only when it returns, so anything checking "do we have a session?" during that round trip starts a second one. The screen re-arms a 1.2s debounced `maybeStartSession` on every rebuild, so a rebuild landing in that window did exactly that. It only shows on an install with no stored session, which is why it surfaced during testing on a fresh install. Fixed in e82fab0e by joining an in-flight start, with a test that holds the create call open; without the guard it opens three sessions.

Verified on a cleared Android 9 install: one greeting where there were two.

kar:
On the reported keyboard stutter: measured on a OnePlus 7T (Android 12) with a scripted four-toggle run on the chat screen, three runs per build, via `dumpsys gfxinfo`.

| build | janky frames | 90th percentile |
| --- | --- | --- |
| main, debug | 26.6% / 28.0% / 27.5% | 57-69ms |
| this branch, debug | 28.4% / 30.5% / 32.3% | 61-65ms |

Indistinguishable, so the animation is equally janky without this branch. The comparison that made it look like a regression was against the release build that was installed on the device beforehand; a debug build renders that animation far worse.

Even so, c601c285 scopes the window changes to Android 10 and below, where the bug actually lives. Android 11+ keeps FLAG_LAYOUT_NO_LIMITS and the default soft-input mode, so those devices are byte-identical to before this branch (confirmed via `dumpsys window`: the flag is back on Android 12). Android 9 re-verified: composer still above the keyboard.

kar:
Some jank I experience but it may be my old oneplus thing, and also not related to this PR. I'll gather feedback on it later on other devices and address if needed.

PR #1259, opened 2026-09-15
ci(ios): clean stale Flutter build state before the iOS build

kar:
## Problem

The **iOS build** job runs on a persistent self-hosted runner whose `_work` checkout is reused between runs, and the job had no clean step. Stale `common/build` artifacts caused spurious failures at the framework-packaging step:

```
├─Building App.xcframework and native assets...   ✓
├─Building CocoaPod frameworks...                 BUILD FAILED (~1.6s)
Unable to build CocoaPod plugin frameworks: ** BUILD FAILED **
```

`App.xcframework` (the Dart code) built fine — only the native plugin-frameworks packaging died, fast, which is corrupted/stale build state rather than a compile error. Because a self-hosted **re-run reuses the same working directory**, re-running failed identically. `SHOW_XCODE_LOG: "0"` also hides the underlying xcodebuild error, so CI only shows the generic "BUILD FAILED".

Verified: a clean-from-scratch build of the same tree (`rm -rf common/build && make -C common build-ios`) **succeeds**, including the exact `Building CocoaPod frameworks` step. So it's stale state, not code or deps (nothing iOS-native merged to main since the last green build).

## Fix

Run `make -C common clean` (`flutter clean`) before the bootstrap step so each iOS build starts from a clean tree. Removing `common/build` is what cleared the fault locally.

Trade-off: forgoes incremental caching on this job (~1–2 min added), which is acceptable on a build-only pool given the reliability win.


kar:
Closing — the iOS CI failure wasn't stale build state after all. Root cause is Xcode 27 rejecting pods with a deployment target below 15.0; the real fix is the generated-Podfile deployment-target patch in fix/ios-podfile-deployment-target-xcode27. A clean step wouldn't have helped (the culprit is .ios/Podfile, regenerated at 13.0 by pub each build).

